PR #41, opened 2026-04-28
Replace JSON-RPC backend with local engine stub

experintellia:
## Summary
Refactored the Remote module to delegate method calls to a new LocalEngine module instead of making JSON-RPC calls to a backend service. This is Phase 2 of a migration to client-side game logic, with LocalEngine providing stub implementations that reject with "NotImplemented" errors.

## Key Changes
- **Remote.js**: Simplified from 158 lines to 47 lines
  - Removed all JSON-RPC plugin integration and network request logic
  - Removed message queue handling
  - Removed endpoint configuration and parameter parsing
  - Now acts as a thin shim that delegates to LocalEngine methods
  - Maintains backward compatibility by accepting (but ignoring) `endpointOverride` and `Remote.NEEDS_QUEUE` arguments
  - Methods still return jQuery Deferred promises to preserve existing `.done()/.fail()` chains in Game.js and bootstrap.js

- **LocalEngine.js**: New module created
  - Defines 19 RPC method stubs (buyPowerup, chargePerp, collectPerp, etc.)
  - Each method returns a rejected promise with "NotImplemented" error message
  - Serves as placeholder for Phase 3 implementation of actual client-side handlers

## Implementation Details
- Remote.addMethod() now validates only the method name and reserved names, delegating actual execution to LocalEngine
- Error handling simplified: methods that don't exist in LocalEngine return a rejected promise
- Backward compatibility preserved for callers in app.js and bootstrap.js that still pass endpoint URLs or Remote.NEEDS_QUEUE
- All 19 game RPC methods are pre-registered in LocalEngine to match the current app.js registration list

https://claude.ai/code/session_01J714SKZ5r2VeCAwBydWC5H

Issue #8, opened 2026-04-28
Phase 2: stub Remote.js to delegate to LocalEngine


experintellia:
Replace JSON-RPC over XHR with thin shims that call into a `LocalEngine` module (Phase 3 fills the engine in).

## Tasks

- [ ] Rewrite `scripts/Remote.js` so:
  - No longer imports `jquery-jsonrpc`.
  - `addMethod(name, endpointOverride)` registers a wrapper `function(...args) { return LocalEngine[name](...args) }` that returns a jQuery `Deferred` (resolve/reject) so existing `.done()/.fail()` chains work unchanged.
  - Preserve `Remote.NEEDS_QUEUE` constant (used at `app.js:142-143`).
  - Initially every `LocalEngine[name]` throws `NotImplemented(name)` — so calls `.fail()` cleanly until Phase 3 populates the engine.
- [ ] Add empty `scripts/LocalEngine.js` that exports an object with stub methods for all 17 RPC names from `app.js:140-159`. Each stub: `return $.Deferred().reject('NotImplemented: ' + name).promise()`.
- [ ] Remove `jquery-jsonrpc` from `bower.json` and `scripts/require.config.js` (or stub require path, lower-risk).

## Acceptance

- No XHR fired during boot.
- `app.remote.X(...)` calls return rejected Deferreds with clear "NotImplemented" reasons.

Depends on: nothing (parallel-safe with the other Phase 2 issues).

experintellia:
Verified complete on master. `Remote.js` has no jquery-jsonrpc; `addMethod` registers wrappers that call `LocalEngine[name]` returning jQuery Deferreds; `Remote.NEEDS_QUEUE = true` preserved. `LocalEngine.js` has all 19 stub methods each rejecting with `"NotImplemented: <name>"`. All acceptance criteria met. Closing as completed by Wave 1.

---
_Wave 2 orchestrator_

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #7, opened 2026-04-28
Phase 2: stub Socket.js as local event bus (no SockJS)


experintellia:
Replace SockJS network plumbing with an in-process event bus that preserves the public interface `app.js` and `Game.js` use.

## Tasks

- [ ] Rewrite `scripts/Socket.js` so the constructor:
  - No longer imports/connects to `sockjs`.
  - Synchronously triggers `connect` then `established` via `$(document).trigger(...)`.
  - Exposes `.on(eventName, handler, needsQueue)` with the existing `Socket.NEEDS_QUEUE` semantics, backed by `RpcQueue`.
  - Exposes `.emit(eventName, data)` that dispatches into the local listener registry (used by LocalEngine in Phase 3 to deliver `node_ready` / `new_items`).
  - `.close()` is a no-op.
- [ ] Remove `sockjs` from `bower.json` and `scripts/require.config.js` (or leave the require path for now and just stub the body — easier rollback).
- [ ] Verify `app.socket.queue.start()` (`Game.js:2063`) still works — the queue infrastructure stays.

## Acceptance

- Page boots without any `sockjs-*` resource being requested.
- `Socket.NEEDS_QUEUE` constant still exported.
- `app.js:266` (`socket.on("node_ready", ..., Socket.NEEDS_QUEUE)`) still parses; events delivered when LocalEngine emits.

Depends on: nothing (parallel-safe with #2-strip-auth and #2-stub-remote).

experintellia:
Verified complete on master. `Socket.js` uses `$(document)` as the in-process event bus — no SockJS. `.on()` with `NEEDS_QUEUE` semantics works, `.emit()` dispatches via `$(document).trigger`, `.close()` removes all subscriptions, `Socket.NEEDS_QUEUE` exported. All acceptance criteria met. (Note: `connect`/`established` are triggered via `setTimeout(0)` rather than synchronously — benign for boot.) Closing as completed by Wave 1.

---
_Wave 2 orchestrator_

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #42, opened 2026-04-28
Replace SockJS with in-process event bus using jQuery

experintellia:
## Summary
Refactored the Socket module to use an in-process event bus backed by jQuery's event system instead of SockJS for WebSocket communication. This change removes the external SockJS dependency and uses the DOM as a message broker for local event distribution.

## Key Changes
- **Removed SockJS dependency**: Eliminated the `require('sockjs')` import and all SockJS-specific configuration (debug, devel, protocols_whitelist, wsUrl)
- **Replaced with jQuery event system**: Socket events now flow through `$(document)` using jQuery's `.on()`, `.off()`, and `.trigger()` methods
- **Simplified event handling**: 
  - Removed the internal `handlers` registry and `trigger()` method
  - Event listeners are now registered directly on the document object
  - Subscriptions are tracked in an array for proper cleanup
- **Updated emit/on semantics**: 
  - `emit()` now triggers jQuery events on the document instead of sending JSON over WebSocket
  - `on()` registers jQuery event listeners that wrap the handler with proper context binding
- **Improved cleanup**: The `close()` method now properly unsubscribes all listeners using `$(document).off()`
- **Automatic connection**: Socket now immediately triggers 'connect' and 'established' events on initialization instead of waiting for WebSocket handshake
- **Code cleanup**: Removed obsolete comments and window beforeunload handler; standardized code formatting

## Implementation Details
- Event data is passed directly through jQuery's event system rather than being wrapped in an envelope structure (ev/pl)
- The queue functionality is preserved for handlers that need it
- All subscriptions are tracked to enable proper cleanup when `close()` is called
- The module maintains backward compatibility with the existing Socket interface while changing the underlying transport mechanism

https://claude.ai/code/session_01LhCMc53Hw4mVgEnV1hdveT

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -20,7 +20,6 @@
     "requirejs-text": "2.0.9",
     "requirejs-tpl-dawsontoth": "https://raw.github.com/dawsontoth/requirejs-tpl/374b6858672cbf32881c950ebe52650dee1b98b3/tpl.js",
     "routie": "0.3.2",
-    "sockjs": "0.3.4",
     "SoundJS": "release_v0.4.1",
     "sprintf": "https://raw.github.com/alexei/sprintf.js/192bc6069ceb0175a98fab67a848568aded77c88/src/sprintf.js",
     "TweenJS": "release_v0.4.1",
--- a/scripts/Socket.js
+++ b/scripts/Socket.js
@@ -1,126 +1,77 @@
-// A tiny wrapper for the SocketIO library, mostly doing some configuration stuff.
+// In-process event bus that mimics the old SockJS-backed Socket interface.
+// Events flow through $(document); the LocalEngine will publish/subscribe on
+// the same channel.
 define(function(require) {
-  
+
   var _ = require('underscore');
   var $ = require('jquery');
-  var SockJS = require('sockjs');
 
   var setup = require('setup');
-  
+
   var Socket = function(path, settings) {
     var self = this;
-    
-    // Reorder the arguments if no path is given.
+
     if (arguments.length < 2) {
-      settings = path
-      path = ''
+      settings = path;
+      path = '';
     }
-    
-    // Merge setup with the settings argument.
+
     settings = _.extend({}, setup, settings);
-    
-    // Options are directly defined from the global setup but can be overridden by the `settings` argument.
-    var options = {
-      debug: settings.debug,
-      devel: settings.wsDevMode,
-      protocols_whitelist: settings.wsProtocolsWhitelist
-    };
 
-    // Connect to the resulting URL and retrieve a socket object.
-    // Also expose the underlying socket object.
-    var socket = this.__socket__ = new SockJS(settings.wsUrl + path, null, options);
-    
-    // The SockJS handlers to delegate messages to our own handlers.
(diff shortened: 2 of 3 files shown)


experintellia:
## Wave 1 Orchestrator Review — Issue #7 (stub Socket.js) ❌ CHANGES REQUESTED

The SockJS removal and jQuery event-bus wiring are directionally correct, but there are two blockers.

---

### Blocker 1 — `connect` and `established` fired before handlers are registered

**Location:** bottom of the Socket constructor, just before `return this`:
```js
$(document).trigger('connect');
$(document).trigger('established');
```

These fire **synchronously during `new Socket(...)`**. At that point `app.js` has not yet called `.on('connect', ...)` or `.on('established', ...)` — those register after the constructor returns. jQuery document triggers are synchronous: the events fire into the void.

**Consequence:**
- `connect` handler (app.js:234) never sends `client_connected`.
- `established` handler (app.js:241) never resolves the `handshake` deferred → `loadGame` (app.js:178, gated on `handshake`) never runs. **App hangs silently.**

**Fix:** defer by one tick so `.on()` calls are registered first:
```js
setTimeout(function() {
  $(document).trigger('connect');
  $(document).trigger('established');
}, 0);
```

---

### Blocker 2 — `sockjs` not removed from `bower.json` / `require.config.js`

`require('sockjs')` is gone from the module body ✓, but `sockjs` almost certainly still appears in `bower.json` (dependency entry) and `scripts/require.config.js` (paths/shim entry). If those remain, the build still downloads and bundles the library — the point of this issue.

Review checklist: *"is the dep removed in ALL its homes (bower.json, require.config.js, the module body, any callers)?"*

**Fix:** remove the `sockjs` entry from `bower.json` and its `paths`/`shim` entry in `scripts/require.config.js`.

---

### Merge-order note (for when you resubmit)

PR #41 (issue #8) also needs to touch `bower.json` and `require.config.js` to remove `jquery-jsonrpc`. Recommended order is **#7 first, then #8**. Whichever lands second will need a small rebase on those two files. Once both are fixed and ready, I'll post the conflict note on the second one.

---

**Non-blocking observations:**
- Removing `this.trigger()` as a public method is fine — no external callers in `scripts/`.
- Removing `beforeunload` is appropriate — no real socket to close.
- `$(document)` shared namespace is fine for MVP (single Socket instance per page).

---

experintellia:
## Wave 1 Orchestrator Re-review — Issue #7 (stub Socket.js) ✅ LGTM

Both blockers resolved:

1. **`connect`/`established` timing** — wrapped in `setTimeout(..., 0)` ✓. Handlers registered by `app.js` will be in place before the events fire.
2. **`sockjs` dep fully removed** — `bower.json` entry gone ✓, `require.config.js` `paths` entry and `shim` block both gone ✓.

**Merge-order note:** PR #41 also touches `bower.json` and `require.config.js`, but on different lines (removing `jquery-jsonrpc`/`jquery-message-queuing`). The two branches will auto-merge cleanly — no rebase needed for whichever lands second.

**LGTM — @experintellia ready to merge as order 7 of 8.**

---

PR #112, opened 2026-05-01
Fix #103: weighted-average share merge for Database token bars

### LINKED ISSUE #103: Restore weighted-average share merge for TokenPerp database bars (follow-up to #98)
Follow-up to investigation #98 (PR #102, `docs/ui-meters.md`).

## Context

Both UI meters flagged in #98 are bound to the right state field; the
LocalEngine port computes the *value* with the wrong arithmetic.
`instance_data.amount` on a `TokenPerp` (called `share` upstream) is
documented in `dd_app/dd_calc.py` `Database.merge` as **the
weighted-average prevalence (0–100 %) of that token type across every
profile in the player's database**:

```
new_share = min(100,
  ( db_share * (M - dupes)
    + dupe_count * dupe_mix
    + ps_share * N
  ) / (M + N))
```

where `M` = `profiles_value` *before* the merge, `N` = the new
profileset's `profiles_value`. In particular, tokens the new
profileset does *not* contain are **diluted** to
`db_share * M / (M + N)`, which is what makes the bar move down as
well as up.

`scripts/LocalEngine.js:1709-1711` instead clamp-sums per-integration
deltas:

```js
var newAmount = Math.min(100,
  ((n.instance_data && n.instance_data.amount) || 0) + (tok.amount || 0));
```

Combined with a ruleset where every `tokens[].amount === 100`, one
integration of any contact saturates every token in its `tokens` list
to 100 % and the bar (and the derived `crosssum` orange status bar +
% label) reads full forever.

## Proposed fix

In `scripts/LocalEngine.js` `integrateCollected` /
`_integrateProfiles`:

1. Capture `M = state.game_values.profiles_value || 0` *before* the
   `profiles_value` increment.
2. For every existing `TokenPerp` node in `nodes`:
   - if `tokensMap[gestalt]` is present:
     `new_share = min(100, (old_share * M + ps_share * N) / (M + N))`
   - else (token is in DB but absent from this profileset):
     `new_share = min(100, old_share * M / (M + N))` *(dilution)*
3. For first-time-integrated gestalts (newly seeded `TokenPerp` node):
   `new_share = min(100, ps_share * N / (M + N))` — collapses to
   `ps_share` when `M = 0`.
4. Skip the dupe term (port has no per-profile dupe tracking — the
   `integrated_ids[collectId]` guard already short-circuits a re-run
   of the same profileset).

Estimated diff: ~10 LOC in `LocalEngine.js`.

## Knock-on work

- **Tests.** `tests/handlers/collect-integrate.test.js` currently
  asserts `instance_data.amount === 5` after `2 + 3` and `=== 100`
  after `90 + 50`. Re-derive expected values from the new formula and
  update the four affected specs (`increments token node`, `caps at
  100`, `caps a fresh-seeded TokenPerp`, `appends a new TokenPerp
  node`). Add at least one new spec asserting **dilution** of an
  un-touched token after a merge.
- **Mission progression.** `goal.current_amount =
  groot.DBTokensAbsolute[goal.target] || 0` — with shares no longer
  pinned at 100, `mission002` (target 900 of `token008`) no longer
  completes after one integration. Two paths:
  1. Lower the mission target so the corrected curve still hits it
     within the early game (cheapest, ruleset-only change).
  2. Accept the longer pacing as part of the original game's
     intended difficulty (no ruleset change; document in
     `docs/ui-meters.md`).
  Recommend (1) for first-playable; (2) eventually.
- **Ruleset balance review.** Every `tokens[].amount` in
  `data/ruleset_3.{en,de}.json` is currently `100`. Once the merge is
  weighted-averaged, the steady-state share of a token is determined
  by *which* contacts the player owns and how often each is
  collected. Play-test to confirm none of the existing missions
  become unwinnable; rebalance individual `tokens[].amount` rows if
  needed (these are vendored from `dd_rules` — coordinate with the
  rules repo per `data/UPSTREAM.txt`).

## Acceptance

- [ ] `_integrateProfiles` switched to weighted-average merge with
      dilution of untouched tokens.
- [ ] `collect-integrate.test.js` specs updated; new dilution spec
      added; `pnpm test` green.
- [ ] Per-TokenPerp database bars visibly move both up and down across
      successive integrations.
- [ ] Status orange bar / % label tracks the same and is no longer
      pinned at 100 %.
- [ ] All existing missions remain winnable (target rebalanced if not).

## References

- `docs/ui-meters.md` (introduced in PR #102) — full investigation
  output and binding map.
- `dd_app/dd_calc.py` `Database.merge`, `dd_app/dd_merger.py`
  `UpgradeToken/Merger`, `dd_app/chargecollect.py`
  `CollectableToken.getPerpChargeData` — upstream merge math.
- #98 — investigation issue (closed by PR #102).
- #81 (round 7) — introduced the clamp-summing semantic when the
  raw-count overflow was patched.

experintellia:
Closes #103. Follow-up to the investigation merged in #102.

## Summary

`integrateCollected` now merges ProfileSet shares with the upstream
`dd_app/dd_calc.py:Database.merge` formula instead of clamp-summing
per-integration deltas. Same math as the server used; just running on
the client because there is no server in the webxdc port.

For every TokenPerp the new profileset *touches*:

```
new_share = min(100, (db_share·M + ps_share·N) / (M + N))
```

For TokenPerps the profileset does *not* contribute to:

```
new_share = (db_share·M) / (M + N)
```

…where `M` = `profiles_value` *before* the merge, `N` =
`ps.profiles_value` (= `increment`, so 0 on duplicate `collect_id`
replay → shares unchanged). First-time-seeded TokenPerps use the same
formula, which collapses to `ps_share` when `M = 0`.

The dilution case preserves the absolute count
(`profiles_value × share / 100`), so
`mission_goals.current_amount` — clamped monotonically in
`_advanceIntegrateProfilesMissions` — hits the same inflection points
the original game did. **No mission targets, no ruleset rows, and no
ruleset balancing values were touched.** Stays faithful to original
gameplay/progression — only the transport changed (LocalEngine on the
client instead of the Python backend).

## Files

- `scripts/LocalEngine.js` — `_integrateProfiles` loop replaced with
  the weighted-average merge + dilution + duplicate-replay no-op.
  Comment block transcribes the upstream formula.
- `tests/handlers/collect-integrate.test.js` — three pre-existing
  share assertions re-derived from the new formula; new specs added
  for **dilution** of an untouched token (the whole point of the
  fix) and **duplicate-replay no-op** (`N = 0`). Mission-progression
  specs pass unchanged because absolute counts are preserved.
- `tests/e2e/share-merge.spec.ts` (new) — Playwright spec drives
  `LocalEngine.integrateCollected` directly (matches
  `gameplay.spec.ts`'s no-canvas-clicks strategy) through two crafted
  db_queue entries, asserts dilution of an absent token, that the
  status-bar `crosssum` drops well below 100 % once shares diverge,
  and that a duplicate `collect_id` replay is a no-op.
- `docs/ui-meters.md` — updated in-place: the previous "non-trivial
  follow-up" section is rewritten as a closed-out trail of the three
  concerns that were resolved (re-derived fixtures, missions still
  hit their targets, ruleset shape vendored unchanged).

## Test plan

- [x] `pnpm test` → 387/387 vitest pass.
- [x] `pnpm test:e2e` → 7/7 Playwright pass (boot ×2,
      buy-perp-error, gameplay, persistence, share-merge ×2).
- [x] `pnpm build` clean.
- [x] Mission-progression specs (`integrate_profiles flow`,
      `monotonic`) pass unchanged — confirms preserved absolute count.
- [ ] Reviewer to spot-check the formula transcription against
      `dd_app/dd_calc.py:Database.merge` (linked in
      `docs/ui-meters.md`).

cc @experintellia for review.

https://claude.ai/code/session_threadHH_103

---
_Generated by [Claude Code](https://claude.ai/code/session_01SDrJ596bBmy2oWLN29STdw)_

### DIFF
--- a/docs/ui-meters.md
+++ b/docs/ui-meters.md
@@ -63,9 +63,9 @@ one is `100`). Tokens that the new profileset does *not* contain get diluted:
 So the bar was an honest "how saturated is this trait in your DB" indicator
 that drifted up and down as new (possibly off-topic) profilesets were merged.
 
-### What the port does instead
+### What the port did before #103
 
-`scripts/LocalEngine.js:1709-1711` (round 7 of #81) clamp-sums:
+`scripts/LocalEngine.js` (round 7 of #81) clamp-summed:
 
 ```js
 var newAmount = Math.min(100,
@@ -73,18 +73,30 @@ var newAmount = Math.min(100,
 ```
 
 Combined with the ruleset where every `tokens[].amount === 100`, *one*
-integration of any contact saturates every token in its `tokens` list to
-100 %. The bar is full forever after.
+integration of any contact saturated every token in its `tokens` list to
+100 %. The bar was full forever after.
 
-### Where the LocalEngine port should bind / compute
+### What the port does now (#103)
 
-The bar binding is fine (`instance_data.amount` ↔ `database_amount` ↔ bar
-fill); the *computation* in `_integrateProfiles` needs to switch from
-`min(100, old + new)` to the weighted-average merge above. New token nodes
-seeded for first-time integration need the same formula, which collapses
-correctly when `M = 0` to `new_share`.
+`integrateCollected` runs the upstream weighted-average merge:
 
-This is **non-trivial** — see the *Why this is not a small fix* section below.
+```js
+var M = (state.game_values.profiles_value) || 0;   // BEFORE the merge
+var N = increment;                                  // 0 on duplicate replay
+// per existing TokenPerp node:
+//   tok present:  newShare = min(100, (oldShare*M + tok.amount*N) / (M+N))
+//   tok absent:   newShare =          oldShare*M / (M+N)
+// per first-time-seeded TokenPerp:
+//   newShare = min(100, tok.amount * N / (M+N))    (= tok.amount when M = 0)
+```
+
+Dilution preserves the absolute count `profiles_value × share / 100`, so
+`mission_goals.current_amount` — fed by absolute counts and clamped
+monotonically in `_advanceIntegrateProfilesMissions` — keeps advancing.
+No ruleset changes; no mission-target rebalancing. The math is a verbatim
+transcription of `dd_app/dd_calc.py:Database.merge` running on the client
+instead of on the server, in line with the wider "port faithfully, just
+drop the server" project goal.
 
 ---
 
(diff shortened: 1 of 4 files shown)


experintellia:
**LGTM — closes the #98 → #102 → #103 trail. Ready to merge (Group A).**

This is the textbook follow-up to #102's investigation. Per the original orchestrator rule for #98:

> ACCEPT a docs-only PR that documents findings and either restores the bindings if the original meaning was recoverable, OR files a follow-up issue if the meaning is unrecoverable + proposes a clearly-defined replacement. REJECT a code-fix PR that "guesses" at meter semantics without source-document evidence.

The original meaning was recoverable (weighted-average share, documented in #102 with citations to `dd_app/dd_calc.py:Database.merge`); this PR is the recovery. Formula is transcribed verbatim from the upstream source the investigation cited — not guessed.

## Spot-checks against the formula

- ✅ Token in PS: `(oldShare * M + tok.amount * N) / (M+N)` — implementation `psContrib = tok ? (tok.amount||0) * N : 0` then `Math.min(100, (oldShare*M + psContrib) / denom)`.
- ✅ Token not in PS (dilution): `psContrib = 0` → `oldShare * M / (M+N)`. Always ≤ oldShare for N>0, so the `Math.min(100,…)` clamp is a no-op here. New `'dilutes a TokenPerp not present in the new profileset'` test asserts both the share value (40) and the absolute-count preservation invariant.
- ✅ First-time seed: `seedShare = Math.min(100, tok.amount * N / denom)` collapses to `tok.amount` when M=0. Existing fresh-seed test still passes.
- ✅ Duplicate `collect_id` replay (N=0): `Math.min(100, oldShare*M / M) = oldShare` for every existing node, and the seeding loop bails because `denom !== 0` but the seed share rounds to 0. New `'does not change shares on a duplicate collect_id replay'` unit test + the e2e `'integrating the same profileset twice does not change shares'` test cover both branches.
- ✅ `denom === 0` bail (skipMerge): defensive; only reachable on M=0 + N=0 corner. Comment explains the case.
- ✅ Non-TokenPerp guard: `if (n.game_type !== 'TokenPerp' || !n.gestalt) return n;` — ContactPerp/ClientPerp/ProjectPerp untouched.
- ✅ Absolute-count preservation under dilution: `(M+N) × (oldShare*M/(M+N)) / 100 = M × oldShare / 100`. Confirmed both algebraically and by the new test's `expect(absAfter).toBeCloseTo(absBefore, 6)`.

## Mission progression invariant

The "mission goals still race ahead" claim in the doc update checks out: `groot.DBTokensAbsolute[goal.target] = profiles_value * amount / 100` is preserved under dilution, and tokens that *every* profileset contributes to (like `token008` per ruleset) stay at 100 % — so `mission002` still completes when `profiles_value ≥ 900` exactly as before. Existing `mission progression — integrate_profiles flow` specs pass unchanged, which is the strongest evidence.

## Architecture invariants

- ✅ Pure function over state; no `webxdc.sendUpdate` site added; not inside `applyDelta` (still inside `integrateCollected` action site).
- ✅ ESM, no AMD `define`, no `setTimeout` for game cycles, no DOM globals in handler bodies.
- ✅ Replay-safe by construction: N=0 dup-collect replay is a no-op (verified).
- ✅ No ruleset rows touched, no mission targets rebalanced — fidelity to original game preserved as the doc claims.

## Doc update

The in-place rewrite of `docs/ui-meters.md` is the right call — keeps the trail readable for future archeologists ("What the port did before #103" → "What the port does now (#103)") rather than spawning a parallel "fix log" file.

## Cross-PR coordination

No conflict with the multi-device-sync work (#114/#115/#117 just filed) — those touch reducers in `state.js`, this touches `_integrateProfiles` math in `LocalEngine.js`. Good to merge ahead of any of those.

@experintellia — clean follow-up, source-cited, comprehensive tests. Group A, ready to merge anytime.

---

Issue #114, opened 2026-05-01
collectPerp reducer leaves orphan nodes_charging entry → perp collectable again after reload


experintellia:
## Bug

After collecting from a perp and reloading the page, the perp sometimes appears collectable again — the user can collect a second time without a charge in between. Reproduced during the #28 manual smoke test.

## Root cause

`scripts/state.js:346-384` — the `collectPerp` reducer removes the path from `nodes_collect` but **never touches `nodes_charging`**. During live operation this is harmless because `LocalEngine.collectPerp` calls `materialize()` first (which strips the charging entry in-memory) before committing the delta. But on **replay-from-zero** at boot:

1. `chargePerp` delta replays → adds entry to `nodes_charging` with `charge_end`.
2. `collectPerp` delta replays → strips `nodes_collect`, leaves `nodes_charging` intact.
3. Next `materialize()` call sees `now ≥ charge_end` and re-promotes the entry to `nodes_collect`.
4. UI shows it as collectable again. User clicks. Free cash / profiles.

## Fix sketch

In the `collectPerp` reducer in `scripts/state.js`, also filter `nodes_charging` by `path` before returning, so the reducer is self-sufficient and idempotent under replay.

```js
var newCharging = (state.nodes_charging || []).filter(function (e) {
  return e.path !== path;
});
// include nodes_charging: newCharging in the returned state
```

## Acceptance

- New `tests/e2e/collect-after-reload.spec.ts` Playwright test reproduces the bug today and passes after the fix. Outline:
  1. Buy + charge a perp (e.g. `contact035`).
  2. Advance `window.__dd.advanceNow(31_000)` past `charge_end`.
  3. Click `[data-testid="dd-collect-button"]`; assert collect-ready decorator gone.
  4. `await page.reload()`.
  5. Assert `[data-testid="dd-collect-ready"]` for the path is **not** present, and cash is unchanged from step 3.
- Existing unit tests in `tests/handlers/collect-integrate.test.js` extended with a replay-from-zero case: apply `chargePerp` + `collectPerp` deltas to a fresh state, materialize, assert the path is not in `nodes_collect`.

## Notes

This is the same anti-pattern flagged in #98's investigation and the multi-device-sync audit: the in-memory materializer state was treated as canonical post-collect, but only the delta is persisted. See sibling issue for the broader audit ask.

Issue #117, opened 2026-05-01
Ensure state.addr is set before any applyDelta call (boot ordering)


experintellia:
## Bug

`scripts/state.js:494` — the `applyDelta` addr guard `if (state.addr && delta.addr && delta.addr !== state.addr)` silently drops deltas when `state.addr` is falsy (typically because boot hasn't set `state.addr` yet). At boot, before `webxdc.selfAddr` is propagated into `state.addr`, any early-arriving delta replay is silently lost.

In practice this manifests as: brand-new `.xdc` open, the very first delta in history might not apply correctly because `state.addr` is `null`/empty when replay starts.

## Fix

Guarantee `state.addr = webxdc.selfAddr` is set **before** the first `applyDelta` call (whether from a live handler emit, the `setUpdateListener` callback, or boot-time replay). Walk `scripts/boot.js`:

1. `webxdc.selfAddr` is available as soon as `webxdc.js` is loaded — that's earlier than any handler can fire.
2. Initialize `state.addr` from `webxdc.selfAddr` as the first thing after `freshState()`, before `setUpdateListener` is registered or any historical delta is replayed.
3. Make `state.addr` set unconditional on the first applicable delta path so we cannot regress.

Optional hardening (separate change, not blocking this one): make `applyDelta` log a warning rather than silently skip when `delta.addr` is empty — silent drops hide bugs.

## Acceptance

- New unit test in `tests/state.test.js` (or wherever `applyDelta` is tested): boot with a non-empty history; assert `state.addr === webxdc.selfAddr` before any delta is applied; assert all historical deltas are reflected post-replay.
- Manual: open a fresh `.xdc`, immediately reload, confirm no state is lost in the second boot.

## Severity

P2 — easy to mis-trigger in dev because the mock `webxdc.selfAddr` is set synchronously, but real Delta Chat may have any timing. Worth fixing before Phase 6 (#29/#30 leaderboard) lands, since that's the work that actually depends on `state.addr` being correct from the very first delta.

Issue #116, opened 2026-05-01
Drop async wrappers around app.remote.* — handlers run in-process, no server


experintellia:
## Context

`app.remote.*` returns Promises (was `$.Deferred` originally) because the original game shipped a JSON-RPC backend that called a real server over SockJS. After the webxdc port (#41 + #42), `app.remote.*` is a thin wrapper around synchronous `LocalEngine` calls that all run in-process — there's no server, no network, no waiting.

The async wrapper still leaks into call sites in `scripts/Game.js`:

- Optimistic UI mutation patterns: handlers mutate `groot.raw_data.*` synchronously, then `app.remote.*().done(...)` for the actual commit. Examples: `dismissMissionBriefing` (Game.js:3008-3014), `markTokenSeen` (Game.js:3051-3056), and others.
- Race window: if the app closes between the optimistic mutation and the RPC commit, the UI state mutates but no delta is broadcast — peers never see it.
- Reasoning overhead: every handler has a `.fail()` reconciliation path that's dead code (the local handler can never actually fail asynchronously).

## What to do

1. Make `app.remote.*` return synchronously (not a Promise). Either drop the wrapper entirely and call `LocalEngine` directly, or keep the wrapper as a sync passthrough for AMD compatibility (Game.js calls `app.remote.*` from AMD code; LocalEngine is ESM).
2. Inline the optimistic-then-commit pattern at every call site: just call the handler and let the commit be the only mutation site. No more `groot.raw_data.* = ...` before the call.
3. Drop the `.done(...)` / `.fail(...)` continuations.
4. Update tests to match.

## Acceptance

- `app.remote.*` is sync (or the AMD wrapper that wraps `LocalEngine` is sync).
- No call site in `scripts/Game.js` mutates `groot.raw_data.*` before calling the handler — the handler write is the only mutation.
- All existing tests pass; race-window tests removed if any (no longer applicable).
- One follow-up smoke pass on the local single-player loop (charge / collect / integrate / dismiss-briefing) to confirm UI feedback still feels responsive.

## Out of scope

- Don't change `LocalEngine` itself; it's already sync internally.
- Don't change the multi-device sync semantics — `webxdc.sendUpdate` is still fire-and-forget at the action site.

## Notes

This cleanup is partially MVP-blocking via #28: the `dismissMissionBriefing` race window means a peer that dismisses on phone and immediately closes might not have broadcast the delta to desktop. Issue is small and self-contained.

PR #119, opened 2026-05-01
Fix chargePerp double-deducting cash/AP on listener echo

### LINKED ISSUE #115: Audit all state.js reducers for replay-safety + idempotency under self-echo
## Context

Sibling of the `collectPerp` replay-safety bug. That bug is one instance of a wider pattern: handlers in `scripts/LocalEngine.js` materialize state in-memory and commit a delta whose reducer in `scripts/state.js` doesn't reproduce the materialized cleanup on its own. Live operation looks fine because the optimistic in-memory state path is canonical between commit and reload; replay-from-zero at boot is where it breaks.

## What to audit

For every state-changing handler in `scripts/LocalEngine.js`, check the corresponding reducer in `scripts/state.js`:

- Does applying the delta to a fresh state (no materialization) produce the exact same state shape as the live handler produces post-commit?
- Are all source collections (`nodes_charging`, `nodes_collect`, `db_queue`, `mission_goals`, `active_missions`, `nodes`) updated consistently?
- Is the reducer **idempotent** — applying the same delta twice (because of self-echo + replay) is a no-op?

Handlers to walk through:

- `buyPerp` / `chargePerp` / `collectPerp` / `integrateCollected`
- `buyKarma` / `buyPowerup` / `sellPowerup` / `buySlots`
- `setDisplayName` / `setPerpCoordinates`
- `dismissMissionBriefing` / `markTokenSeen`

For each, check both the live `_commitDelta` path **and** the replay path, and confirm they end at the same state.

## Output

A short audit report (one paragraph per handler) under `docs/replay-safety-audit.md`, plus follow-up bug issues for any handler that fails the test. Existing tests in `tests/handlers/*.test.js` mostly cover the live path; add a replay-from-zero case for each handler that's missing one.

## Acceptance

- Audit doc landed.
- Per-handler replay-from-zero unit tests in place.
- Any newly-found bugs filed as separate issues with reproduction.

experintellia:
## Summary

Fixes a state-sync bug where the UI shows enough cash/AP to charge a perp but the handler refuses with "no_cash" / "no_AP".

## Root cause

`chargePerp` in `scripts/LocalEngine.js` calls `setState(newStateCharge)` and then emits the delta via `_persistDelta`. In production webxdc, `sendUpdate` echoes the own update back into the listener registered by `boot.js`, which re-applies the `chargePerp` reducer in `scripts/state.js`. That reducer used incremental math (`gv.cash_value - cashDelta`, `ap_snapshot - 1`), so it ran on top of state where the deduction had already been applied — double-charging cash and AP.

The handler returned the once-deducted snapshot to the UI, so the status bar showed values higher than the engine actually held. The gap widens by one `cashDelta` per charge until the engine's actual cash drops below the cost — at which point the next click fails despite a green bar.

Tests didn't catch it because `setSendDelta(...)` short-circuits `_persistDelta` before reaching `webxdc.sendUpdate` (LocalEngine.js:90-92), so the listener never echoed in unit tests.

## Fix

- Carry the post-mutation `game_values` snapshot in the chargePerp delta result.
- Make the reducer prefer the snapshot over the incremental form. Snapshot application is idempotent on its own output, so the listener echo no longer double-deducts.
- Keep the incremental form as a fallback so already-persisted pre-fix deltas still replay correctly on cold start.

## Tests

Three new regression tests in `tests/handlers/chargePerp.test.js` simulate the production webxdc echo by applying the captured delta on top of the post-handler state, asserting cash, cash_spent, ap_snapshot, and xp_value are not double-deducted. Full suite: 391/391 passing.

## Related issues

Addresses one finding from #115 (audit all state.js reducers for replay-safety + idempotency under self-echo) — specifically the chargePerp reducer. Does **not** close #115, which is a broader audit covering every handler.

Related: #114 (sibling instance of the same materializer-vs-reducer anti-pattern in `collectPerp`, with a different symptom — perp re-collectable after reload). Not fixed here.

## Test plan

- [x] Existing unit tests pass (`npx vitest run`)
- [x] New regression tests in `tests/handlers/chargePerp.test.js` cover the listener-echo path
- [ ] Manual smoke in Delta Chat Desktop: charge a perp repeatedly, verify cash and AP in the UI match the engine state across many cycles (no drift)

https://claude.ai/code/session_01Q2ejf3RVooZTdPaFVsDwNB

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q2ejf3RVooZTdPaFVsDwNB)_

### DIFF
--- a/scripts/LocalEngine.js
+++ b/scripts/LocalEngine.js
@@ -1463,13 +1463,21 @@ export function chargePerp(token, path) { // eslint-disable-line no-unused-vars
 
   setState(newStateCharge);
 
+  // Carry the post-mutation game_values snapshot so the reducer is idempotent
+  // when the listener replays this delta on top of the just-setState'd state
+  // (production webxdc echoes own updates into setUpdateListener). Without the
+  // snapshot, the reducer's incremental cash/ap math runs a second time and
+  // double-deducts — the UI then shows cash the engine no longer believes the
+  // player has, so subsequent charges fail with "no_cash" despite a green bar.
+  // cashDelta/xpInc remain in the payload for cold-replay backwards compat
+  // with deltas persisted before this fix.
   _persistDelta({
     kind:   'delta',
     addr:   state.addr,
     op:     'chargePerp',
     args:   [path],
     result: { chargeEntry: chargeEntry, nodeIdx: nodeIdx, cashDelta: chargeCost, xpInc: xpInc,
-              missions: chargeMissionResult.missions || null },
+              game_values: newGv, missions: chargeMissionResult.missions || null },
     ts:     now,
   });
 
--- a/scripts/state.js
+++ b/scripts/state.js
@@ -332,12 +332,19 @@ reducers.chargePerp = function chargePerpReducer(state, delta) {
   });
 
   var gv    = state.game_values || {};
-  var newGv = Object.assign({}, gv, {
-    cash_value:  (gv.cash_value  || 0) - cashDelta,
-    cash_spent:  (gv.cash_spent  || 0) + cashDelta,
-    xp_value:    (gv.xp_value   || 0) + xpInc,
-    ap_snapshot: Math.max(0, (gv.ap_snapshot || 0) - 1),
-  });
+  // Prefer the post-mutation snapshot when the handler ships one — applying a
+  // snapshot is idempotent, so the listener's echo of an own update doesn't
+  // double-deduct cash/ap on top of the handler's setState. Fall back to
+  // the incremental form for replay of older deltas that predate the
+  // snapshot field.
+  var newGv = r.game_values
+    ? Object.assign({}, gv, r.game_values)
+    : Object.assign({}, gv, {
+        cash_value:  (gv.cash_value  || 0) - cashDelta,
+        cash_spent:  (gv.cash_spent  || 0) + cashDelta,
+        xp_value:    (gv.xp_value   || 0) + xpInc,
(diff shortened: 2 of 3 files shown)


experintellia:
**Recommend closing this PR unmerged in favor of #121 (the canonical fix).**

This PR is correct as a tactical patch and CI is green, but the architectural fix in #121 (now LGTM and ready to merge) supersedes it cleanly:

- #121's `chargePerp` reducer change is **identical** to this one — prefer the `r.game_values` snapshot, fall back to incremental for legacy persisted deltas. Same back-compat property: pre-fix deltas in any user's update history still replay correctly on cold start.
- #121 applies the same listener-echo idempotence test pattern that's introduced here, plus the same pattern for **every other game_values-mutating handler** (`buyKarma`, `buyPerp`, `collectPerp`, `integrateCollected`, `buyPowerup`, `sellPowerup`, `buySlots`) — so it structurally rules out the entire bug class, not just the chargePerp instance.
- #121 also closes #114, #115, #116, #117 in the same drop, with a Phase 6 static guard preventing handler-side `setState` regressions.

Merge order:

- **Recommended**: close this unmerged → merge #121 → no churn. The chargePerp change in #121 is a strict superset of this one.
- **Alternative**: merge this first → merge #121 (will require trivial conflict resolution on `scripts/state.js` chargePerp reducer). Functionally equivalent end state, but adds churn for no gain.

Architecturally LGTM as a standalone fix; just no longer needed. @experintellia — your call on which path; I'd close-not-merge.

---

PR #122, opened 2026-05-02
Follow-up to #121: await replay + drive loader progress bar

experintellia:
Follow-up to #121, stacked on its branch. Merge order: #121 → master, then this → master (or rebase on master once #121 lands).

## Why

Per the webxdc spec, `setUpdateListener` returns a Promise that resolves once every update from the registered serial up to the messenger's `max_serial`-at-registration has been delivered to the listener. Today `scripts/boot.js` ignores that promise — `materializer()` and `onReady()` fire immediately, so `app.start()` runs and `Game.loadGame()` reads a partially-replayed state. Not visibly broken on small histories, but a contract violation that grows worse as saves accumulate updates and on slower targets.

## What changes

- **`scripts/boot.js`** — `boot()` now returns a Promise. Idempotent under repeated calls. Tracks per-callback `(serial, max_serial)` on a tiny module-level snapshot and exposes `getReplayProgress()` / `getBootPromise()` for UI polling.
- **`scripts/bootstrap.js`** — requires the AMD-bridged `boot` module, polls `getReplayProgress()` every 100 ms to drive `<progress id="loader">`, and gates `app.start()` on `getBootPromise()` resolution.
- **`scripts/esm-entry.js`** — comment update only; the synchronous `boot()` invocation is unchanged so `getState()` is non-null from the moment AMD modules require `LocalEngine`.

## Edge cases handled

- **No webxdc** (Node/tests): `setUpdateListener` isn't called; `boot()`'s promise resolves immediately.
- **Dev shim** (`scripts/webxdc-shim.js`): returns `undefined` from `setUpdateListener`; `Promise.resolve(undefined)` resolves immediately.
- **Live peer updates mid-replay**: `max_serial` grows beyond the snapshot, so the bar may not visually fill before the listener promise resolves. **Acceptable per design** — the game still starts on time and later peer updates land via the same listener live.
- **Repeated `boot()` calls**: returns the in-flight promise, so esm-entry's invocation and any future explicit re-boot share the same outcome.

## Test plan

- [x] `pnpm test` — 472 passing (no behavioural change for unit tests; they never touch the boot promise path).
- [ ] **Reviewer ask**: manual smoke on Delta Chat Desktop with a save that has non-trivial history (e.g. a few thousand `setPerpCoordinates` deltas accumulated through play). Confirm loader progress bar animates and game UI does not flicker through partial state.
- [ ] Multi-device join: open the same `.xdc` on a second device with active peer chatter. Bar may stop short of full; game should still start.

https://claude.ai/code/session_120_followup_replay

---
_Generated by [Claude Code](https://claude.ai/code/session_01BjWeQ63bs2xq4Bf1QM8Byv)_

### DIFF
--- a/scripts/boot.js
+++ b/scripts/boot.js
@@ -4,7 +4,13 @@
 //   1. state = freshState(selfAddr) — selfAddr seeded BEFORE listener
 //      registration so applyDelta's addr guards never see an empty state.addr
 //      during boot replay (closes #117).
-//   2. webxdc.setUpdateListener(cb, 0)  — replays full update history.
+//   2. webxdc.setUpdateListener(cb, 0) — replays full update history.
+//      The webxdc spec returns a Promise that resolves once every update
+//      from `serial` up to the messenger's max_serial-at-registration has
+//      been delivered to `cb`. boot() awaits that promise so the engine
+//      is fully replayed before materializer / onReady fires; bootstrap.js
+//      can then swap the loader for the game UI without flickering through
+//      a partially-replayed state.
 //      The callback is the SOLE setState site in production: it routes both
 //      own-echoes (from sendUpdate) and remote peer deltas through the same
 //      applyDelta path. Handlers in scripts/LocalEngine.js never call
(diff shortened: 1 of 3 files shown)


experintellia:
**LGTM — clean follow-up on #121. Ready to merge after #121 lands (rebase on master required first).**

This is a real architectural correctness fix, not just polish: the original `boot()` ignored `setUpdateListener`'s return Promise, so `materializer()` and `onReady()` fired before the messenger had finished delivering historical updates. Not visibly broken on small histories but a contract violation that grows worse as saves accumulate. Worth fixing now rather than chasing it under load later.

## What's right

- **`boot()` is now Promise-returning + idempotent.** Subsequent calls return the in-flight promise rather than re-registering the listener — guards against AMD re-`require` and any future explicit re-boot.
- **`Promise.resolve(listenerPromise)` normalizes three shapes uniformly** (real webxdc, dev-shim returning `undefined`, no-webxdc Node). Clean.
- **`_replayProgress` only grows.** Both `serial` and `max_serial` use `if (s > _replayProgress.serial) _replayProgress.serial = s` — monotonic, safe under concurrent peer updates landing mid-replay.
- **`onReplayProgress` callback wrapped in `try/catch`** so a UI-side bug can't break the replay loop.
- **Bootstrap polling at 100 ms** drives the `<progress>` element; cleanup on both `p.done` flag and `continueStart()` paths so the timer can't leak.
- **Edge cases enumerated correctly in PR description.** Live peer updates mid-replay growing `max_serial` is the trickiest one — the bar may not visually fill, but the game still starts at promise resolution. Documented as design intent, not a bug.

## Architecture invariants

- ✅ ESM-friendly; AMD bridge via `esm-entry.js` exposes `boot` to `bootstrap.js` consumers.
- ✅ No new `setState` site (still routes through #121's listener-only rule).
- ✅ Replay-safe by construction: the listener body is unchanged from #121 — it just gets a settle signal now.
- ✅ Module-level state (`_bootPromise`, `_replayProgress`) is fine for a singleton boot module.

## Minor observations (not blockers)

- The `progressTimer` polling at 100 ms is reasonable for a loader UX; if save history grows enormous, a smarter "tick on listener fire" approach (calling the progress callback's UI hook directly) would be tighter, but 100 ms is well below user-perceivable lag.
- `serial: 0, max_serial: 0` initial state has the divide-by-zero/empty-bar guard via `if (p.max_serial > 0)` in bootstrap.js — correct.
- One potential UX wart: if peer updates during replay grow `max_serial` faster than `serial` advances, the bar's fill ratio appears to *regress*. Per design, but worth eyeballing during the manual smoke. The acceptable-end-state of "bar may not fill before game starts" is fine; visible regression mid-replay would be jarring.

## Coordination

- **Stacked on #121**, base = `claude/refactor-architecture-Xafki`. Once #121 lands in master, this needs to rebase.
- **Independent of #123** (AP bar fix + e2e test) — disjoint files; either order works post-#121.
- **No new tests**: vitest suite passes unchanged because none of it touches the boot promise path. Reviewer ask in PR description (manual smoke with non-trivial save history) is the right gate.

@experintellia — Group A, ready to merge after #121 + rebase. Suggested order: **#121 → master**, then **#122/#123 in either order → master**.

---

Issue #133, opened 2026-05-03
Bug tracking: medium/low-severity correctness issues from audit


experintellia:
## Context

Code-quality audit (branch `claude/code-quality-audit-IAHib`) identified a cluster of medium- and lower-severity correctness issues. The four game-breaking ones are filed as their own issues; this is the umbrella for the rest.

> **Status (2026-05-06):** closed. See the [audit comment](#) for per-item resolution. 8 fixed, 1 mitigated, 2 closed as not-reproducible / false-positive.

### Medium

- [x] **Float equality in `integrateCollected`** — fixed in `scripts/LocalEngine.ts:2624` (`Math.abs(newShare - oldShare) < 1e-9`).
- [x] **Zero/negative-denom path in share computation** — fixed in `scripts/LocalEngine.ts:2607-2613` (`Math.max(0, ...)` + `skipMerge` bail).
- [x] **Mission `integrate_profiles` progress can roll backward** — fixed in `scripts/LocalEngine.ts:1679` (`Math.max(prev, computed)`).
- [x] **`clock.advance()` accepts negative deltas** — fixed in `scripts/clock.ts:54` (throws `RangeError`).
- [x] **`_generateId()` collision risk under load** — mitigated in `scripts/LocalEngine.ts:2244-2247` (now two 24-bit RNG words ≈ 48 bits of entropy after the timestamp). Tighten further with `crypto.randomUUID()` only if a real collision shows up.
- [x] **Schema-reset wipes peers with empty addr** — fixed in `scripts/state.ts:783` (reset deferred until `state.addr` is seeded).

### Lower

- [x] **collectPerp de-dup checks wrong field** — closed as not-reproducible. Current `state.ts:576-578` dedupes against `db_queue` by `collect_id`, which is the right field; original audit claim doesn't match current code shape.
- [x] **profilesCollected sum unbounded** — fixed in `scripts/LocalEngine.ts:1711` (clamped at `goal.amount`).
- [x] **buyKarma clamps the increment, not the result** — verified in `scripts/LocalEngine.ts:894`; the clamp wraps the sum, which is arithmetically correct. Audit was a false read.
- [x] **buy_perp mission goal completion doesn't check `active_missions`** — fixed in `scripts/LocalEngine.ts:1955` (and parallel guards added to charge_perp / buy_powerup / upgrade_token at `1872` / `1899` / `1926`).
[...]

PR #138, opened 2026-05-03
Reject negative deltas in clock.advance (#133)

experintellia:
`scripts/clock.js`'s `advance(deltaMs)` did `_override = now() + deltaMs`, so `advance(-1000)` rewound the clock. The reducer's `Math.max(clockNow(), last_seen_ts)` skew guard prevents persisted state from rewinding, but a negative advance can still set up impossible clock states in tests and violates the forward-only contract documented in the file header. `advance()` now throws `RangeError` on a negative `deltaMs` so test bugs surface immediately.

Addresses one item from #133 (does not close it).

A new test in `tests/clock/clock.test.js` covers `advance(-1)` throwing and verifies the clock did not move.

---

### DIFF
--- a/scripts/clock.js
+++ b/scripts/clock.js
@@ -43,7 +43,14 @@ export function clearOverride() {
  *
  * Equivalent to setOverride(now() + deltaMs).  If no override is currently
  * active, the base is the real Date.now() at the moment advance() is called.
+ *
+ * The contract is forward-only: deltaMs must be >= 0.  Passing a negative
+ * value throws RangeError so test bugs surface immediately rather than
+ * silently producing impossible clock states.  (state.js separately guards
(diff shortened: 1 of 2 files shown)


experintellia:
**Review:** ✅ Looks good.

Diff is the minimal correct shape — `RangeError` guard at the top of `advance()`, JSDoc updated, one focused test that asserts both the throw and that the clock didn't move. CI green (unit + e2e).

Optional nit: the error message could echo the bad value (`'clock.advance: deltaMs must be >= 0 (was ' + deltaMs + ')'`) for slightly better stack traces when a test trips it. Not a blocker.

(Cannot use the formal Approve flow because the same account authored the PR.)

---

PR #139, opened 2026-05-03
Stop applyWebxdcIdentity from mutating its argument (#129)

### LINKED ISSUE #129: Bug: webxdc-identity mutates caller's user object
## Severity

Game-breaking (potential state-sync corruption).

## Location

`scripts/webxdc-identity.js:9`

```js
user.display_name = selfName;
```

## Problem

`applyIdentity` (or equivalent) mutates the **passed-in** user object in place rather than returning a fresh copy. Callers in `Game.js` treat the user/raw-data object as immutable; an in-place mutation that bypasses the delta pipeline can:

- corrupt observers that rely on the previous reference being unchanged,
- persist a `display_name` change locally without a corresponding webxdc update, so peers never see the renamed identity,
- leave the player with divergent local-only state that desyncs on the next replay.

## Suggested fix

Return `Object.assign({}, user, { display_name: selfName })` (or use a delta path that flows through the reducer) instead of mutating the argument.

## Acceptance

- [ ] `webxdc-identity.js` no longer assigns into a parameter.
- [ ] Existing `tests/webxdc-identity.test.js` extended with a "does not mutate input" assertion.

Surfaced by code-quality audit (claude/code-quality-audit-IAHib).

experintellia:
Fixes #129.

`applyWebxdcIdentity(user)` previously assigned `user.display_name = selfName` in place. The audit flagged this as a state-sync hazard: callers in `Game.js` rely on `raw_data` references being immutable, so the in-place rename can corrupt observers and persist a local-only rename that never flows through a webxdc delta.

The helper is now non-mutating. Instead of writing into the argument, it returns the new `selfName` string when it differs from the stored one, or `null` when there is nothing to do. The single caller in `Game.js` (the boot path that seeds the display name on first launch) now reads the returned name and dispatches it through `app.remote.setDisplayName(...)`, which is the proper delta path — the `setDisplayName` reducer produces a fresh state via `Object.assign`, so the rename flows through state and observers cleanly.

Test changes: existing tests that asserted the mutated `user.display_name` were updated to assert the returned value instead, and a new test (`does not mutate the input user object`) locks in the non-mutation contract. All 510 tests pass.

---

### DIFF
--- a/scripts/Game.js
+++ b/scripts/Game.js
@@ -2024,7 +2024,7 @@ define(function(require) {
       // so the name survives reloads without prompting the user again. The helper
       // is non-mutating — we route the new name through setDisplayName so the
       // reducer produces a fresh state instead of corrupting the live reference.
-      var newSelfName = webxdcIdentity.applyWebxdcIdentity(this.data.user);
+      var newSelfName = webxdcIdentity.getMessengerDisplayNameChange(this.data.user);
       if (newSelfName) {
         app.remote.setDisplayName(app.token, newSelfName);
       }
--- a/scripts/Game.js
+++ b/scripts/Game.js
@@ -2021,9 +2021,12 @@ define(function(require) {
       this.init(config);
 
       // Seed display_name from webxdc.selfName on first boot; persisted as a delta
-      // so the name survives reloads without prompting the user again.
-      if (webxdcIdentity.applyWebxdcIdentity(this.data.user)) {
-        app.remote.setDisplayName(app.token, this.data.user.display_name);
+      // so the name survives reloads without prompting the user again. The helper
+      // is non-mutating — we route the new name through setDisplayName so the
+      // reducer produces a fresh state instead of corrupting the live reference.
+      var newSelfName = webxdcIdentity.applyWebxdcIdentity(this.data.user);
+      if (newSelfName) {
+        app.remote.setDisplayName(app.token, newSelfName);
       }
 
       this.initGameValues();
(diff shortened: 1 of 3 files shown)


experintellia:
**Review:** ✅ Looks good.

Contract change from `(user) → bool` to `(user) → string|null` is sensible. The call site at `Game.js:2024–2030` correctly routes the returned name through `setDisplayName`, so the reducer is the only writer — no more live-reference mutation. Tests cover all four paths plus the explicit non-mutation assertion and the null-user guard. Unit tests green; e2e was still in progress when reviewed.

Optional nit: the function name `applyWebxdcIdentity` no longer "applies" anything — it only compares and reports. Renaming is out of scope here; flagging for a future cleanup pass.

(Cannot use the formal Approve flow because the same account authored the PR.)

---

experintellia:
Worth noting in passing: display_name can only change in the host messenger — the game doesn't expose any UI to rename. So `applyWebxdcIdentity` is exclusively a "once per boot, did `webxdc.selfName` change since last time?" check, and the dispatched `setDisplayName` only fires on actual messenger-side renames (or first boot).

That tightens the rename consideration from earlier: a future cleanup pass could rename to something like `syncDisplayNameFromMessenger()` or `pickUpRenameFromMessenger()` to make the one-way nature explicit. Out of scope for this PR.

---

experintellia:
Pushed `046d808`: renamed `applyWebxdcIdentity` → `getMessengerDisplayNameChange`. Now matches getter-style of `getToken` / `getSessionLocale` elsewhere and makes the one-way messenger→game direction explicit. Header comment also updated to spell out that the messenger is the only source of name changes. All 510 tests pass.

---

Issue #31, opened 2026-04-28
Phase 2.5: build/test toolchain swap (drop Bower+Grunt, add esbuild+vitest)


experintellia:
**Promoted from Phase 7 to Phase 2.5 — critical-path blocker.** Bower no longer resolves dependencies; Grunt is broken; the existing build path can't produce a runnable artifact.

## Rescoped: toolchain swap ONLY

The original #31 included converting all AMD `define()` modules to ESM. **That part is deferred** to a separate Phase 7 issue, because converting `Game.js` (5669 lines) + `Render.js` (5242 lines) + the rest of the legacy code in the same PR would conflict with every in-flight Wave 2 PR.

**This issue:** swap the *build/test toolchain* without forcing legacy AMD code to change. Make the build runnable again; unblock Wave 2.

## Tasks

- [ ] Delete `Gruntfile.js`, `bower.json`, the `components/` directory.
- [ ] Add `package.json` with **dev-only** dependencies: `esbuild`, `vitest`, `@types/node` if needed. No production deps.
- [ ] Vendor the libs the legacy AMD code actually loads (jquery, underscore, easeljs, tweenjs, preloadjs, soundjs, requirejs, raphael) under `vendor/` as static files. Either npm-install + copy, or download once + commit. Either way the runtime path is `vendor/<lib>.js` with no resolution magic.
- [ ] Configure esbuild to produce a `dist/` bundle:
  - Entry points: index.html + the new ESM modules (LocalEngine, state, materializer, webxdc-shim).
  - Legacy AMD code (`scripts/Game.js`, `scripts/Render.js`, `scripts/app.js`, `scripts/bootstrap.js`, `scripts/Remote.js`, `scripts/Socket.js`, etc.) is loaded by `vendor/requirejs.js` at runtime, **unchanged**.
  - Provide an AMD-compat wrapper so legacy `require('LocalEngine')` resolves to the ESM module's default export. Tiny shim, ~10 lines.
- [ ] Configure vitest:
  - Imports new ESM modules directly from Node (no bundler in test path).
  - Test files live under `tests/` per #45.
- [ ] Rewrite `scripts/build-xdc.sh` (was #26's territory): runs `esbuild` build, then zips the resulting `dist/` + static assets into `data-dealer.xdc`.
- [ ] Verify locally: `pnpm install && pnpm build && pnpm test` all succeed; opening `dist/index.html` in a browser shows the existing "boots-to-broken" state from #9.

## Out of scope (deferred)

[...]

PR #156, opened 2026-05-04
refactor(esm): convert AMD modules to ESM end-to-end (closes #58)

### LINKED ISSUE #58: Phase 7: convert legacy AMD modules to ESM (follow-up to #31)

experintellia:
Closes #58.

Converts every legacy AMD `define()` module under `scripts/` to native ESM and retires the AMD-era plumbing that #31 carried forward (`vendor/requirejs.js`, the `tpl!` plug-in, the `define('name', [], () => __DD[name])` bridge footer).  The codebase now uses **one module system end-to-end**.

Split into five logical commits so each is reviewable on its own; tests, e2e, and `vite build` are green at every commit.

## Commits, in order

1. **`refactor(esm): convert leaf modules (util, setup, i18n, type_settings) to ESM`** — the small modules with no internal-module deps.  i18n drops `$.getJSON` in favour of static `import deAT from '../i18n/de_AT.json' with { type: 'json' }`; setup uses `Object.assign` instead of `_.extend`.  All four are surfaced to the still-AMD callers through the existing bridge footer in `esm-bundle.js`.
2. **`refactor(esm): convert app.js to ESM, expand bootstrap preload list`** — Application factory is now ESM.  `bootstrap.js`'s explicit dep list grows because RequireJS's CommonJS-magic factory string-scan no longer auto-discovers `require('Game')`, `require('numeral')`, etc.
3. **`refactor(esm): convert Game.js (5703 LOC) to ESM`** — wrapper-only port; all ~5670 lines of internal classes byte-identical to master.  Two inner `require('i18n')` / `require('Game')` calls inside the reset/reload callback are now direct module-scope references.
4. **`refactor(esm): convert Render.js (5305 LOC) to ESM`** — same wrapper-only pattern.  No `Game ↔ Render` circular surfaced — the suspected cycle in #58 was a one-way edge `Game → Render → app`, with `app` constructed first, so the back-edge is a no-op.
5. **`chore(esm): retire AMD plumbing — drop requirejs, tpl plugin, AMD bridge`** — bootstrap.js becomes ESM; the `tpl!` plug-in is replaced by `import.meta.glob('../views/*.html?raw', { eager: true })` + `_.template(text, null, { variable: 'D' })` (underscore 1.5.1's signature requires `null` for data to get a precompiled function); vendor libs ship as plain `<script>` tags from `index.html`; `vendor/requirejs.js`, `vendor/almond.js`, `vendor/text.js`, `vendor/tpl.js`, `vendor/native-console.js`, `vendor/jquery-mobile.js` and `scripts/require.config.js` deleted; the `amdBridgeFooter` in `vite.config.js` deleted.

## Things worth flagging for review

- **`window.require` test-access shim.**  The existing playwright e2e specs reach into engine internals via `window.require(['LocalEngine'], cb)` — a shape they inherited from the AMD era.  The cheapest path to keeping them green was a 25-line shim in `esm-entry.js` that exposes `boot`, `LocalEngine`, `app` under fixed names.  This is **not** a generic AMD loader: no `define`, no script fetching, no transitive dep walking.  ~25 LOC vs the ~80 KB requirejs runtime it replaces.  Happy to convert it to a `window.dd` test hook if you'd prefer — would touch all 8 e2e specs.
- **`Render.js` createjs handles.**  The legacy AMD shim returned `createjs` for `createjs-easel` / `createjs-sound` but `createjs.Tween` for `createjs-tween`.  My first pass aliased all three to `globalThis.createjs`, which regressed `Tween.hasActiveTweens(...)`.  Fixed by setting `Tween = globalThis.createjs.Tween` explicitly.  Caught by `gameplay.spec.ts`.
- **`bootstrap.js` Node-import guard.**  Bootstrap's top-level body renders the loader template into the DOM.  In vitest there is no DOM, so I wrapped the body in `if (typeof window === 'undefined' || typeof document === 'undefined') { /* skip */ } else { runBrowserBootstrap() }`.  Keeps the module import-safe without forcing a jsdom dependency.
- **`vendor-install.js` rewrite.**  The CLI tool that regenerates `vendor/` no longer installs requirejs, almond, text, tpl, native-console or jquery-mobile.  If anyone re-runs it, they get only the libs `index.html` references.

## Architecture invariants

None violated.  Reducer logic, `applyDelta` guards, the `addr === state.addr` per-self check, the no-`setTimeout`-for-game-cycles rule, and the `webxdc.sendUpdate({info})`-only-at-original-action-sites rule are all untouched — this PR is purely a module-system migration.

## Acceptance grep

```
$ grep -rn "define(" scripts/                                    # nothing
$ grep -rn "require\.config\|requirejs" scripts/ index.html      # nothing
$ ls vendor/requirejs.js scripts/require.config.js               # No such file
```

## Bundle size

| artifact                         | before  | after   |
| -------------------------------- | ------- | ------- |
| `data-dealer-hq.xdc`             | 14.2 MB | 14.1 MB |
| `data-dealer-casual.xdc`         | 5.5 MB  | 5.4 MB  |
| `dist/scripts/esm-bundle.js`     | ~830 KB | ~1.05 MB |

The bundle itself grew because every view template is now inlined, but the .xdc shrunk because `vendor/requirejs.js` (~80 KB) + `vendor/almond.js` + `vendor/text.js` + `vendor/tpl.js` + `vendor/native-console.js` + `vendor/jquery-mobile.js` + `scripts/require.config.js` + `scripts/bootstrap.js` + `scripts/app.js` + raw copies of `Game.js` / `Render.js` + the `views/` directory are all gone from `dist/`.

## Tests

- `npm test` — **615/615 vitest** (15 new negative assertions for the deleted AMD plumbing).
- `npx playwright test` — **13/13 chromium e2e**.
- `npm run build` — produces working `data-dealer-hq.xdc` + `data-dealer-casual.xdc`.

https://claude.ai/code/session_01DAjKygDMuwtPmG7WuEn7TB

---
_Generated by [Claude Code](https://claude.ai/code/session_01DAjKygDMuwtPmG7WuEn7TB)_


experintellia (commented):
**Review verdict: APPROVE (posted as comment because GitHub blocks self-approval).**

Reviewed commit-by-commit. All five pass the Wave 7b architecture invariants and the per-commit checklist; the engine layer is byte-identical between PR base and head.

---

### Cross-cutting checks (all green)

- **Engine layer untouched.** `git diff e3ddbd5b..571c7688 -- scripts/LocalEngine.js scripts/state.ts scripts/materializer.ts scripts/clock.ts` returns empty. Invariant 3 satisfied at the file level — no reducer, state-shape, handler-logic, or pure-function edits slipped in.
- **`addr === state.addr` guard intact.** `scripts/state.ts:750` still reads `if (delta.addr && delta.addr !== state.addr) { ... }`. Invariant 9 (and the earlier per-self-reducer guards on `mission_briefings_seen` / `tokens_seen`) preserved.
- **No new `webxdc.sendUpdate` emit sites.** Grep on the PR head shows only the two pre-existing emit sites in `scripts/LocalEngine.js`: line 151 (`triggerAchievement`, Wave 7a) and line 165 (`_persistDelta` Node fallback). Both are at original action sites; ESM conversion did not introduce new emit sites. Invariant 6 satisfied.
- **No `setTimeout` for game cycles.** The only `setTimeout` in the diff is in `scripts/util.js`'s pre-existing `wait()` helper (carried byte-identical, just rewrapped from AMD to ESM). Invariant 1 satisfied.
- **AMD-survivor grep on the head:** `grep -rn 'define(\|require\.config\|requirejs' scripts/ index.html` on `origin/claude/phase7-amd-to-esm-Xard8` returns **zero matches**. (Only `vendor/sprintf.js` retains an anonymous `define()`, but it's a vendor file outside `scripts/`, and its no-op when no AMD loader is present is documented in `app.js`.)
- **Vendor AMD loaders deleted.** `vendor/requirejs.js` (-2145), `vendor/almond.js` (-405), `vendor/text.js` (-412), `vendor/tpl.js` (-30), `vendor/native-console.js` (-6), `vendor/jquery-mobile.js` (-15), `scripts/require.config.js` (-68) — all gone from the tree on the head ref.

---

### Per-commit verdicts

**Commit 1 — `ac18943` leaves (util, setup, i18n, type_settings, setup_local).** APPROVE.
- Pure ESM; no `define(` survivors in the touched files.
- Default-vs-named exports chosen sensibly: `i18n` and `setup` stay default (single-object surface), `type_settings` exports the named `getTypeSettings` plus a default for compat, `util` exports both `extend`/`wait` named and a default object.
- `i18n.js` drops `$.getJSON` for static `import deAT from '../i18n/de_AT.json' with { type: 'json' }` — eager and synchronous, which matches the new boot order. `ready()` retained as `Promise.resolve(i18n)` to keep legacy `.then()` callers compiling.
- `setup.js` drops `_.extend` for `Object.assign` — kills the underscore dependency at module load time, which is what allows this file to be imported eagerly inside the bundle IIFE before vendor scripts run.
- `type_settings.js` and `util.js` correctly defer `globalThis._` / `globalThis.jQuery` reads to call time, not module-evaluation time. Comments explicitly call out the ordering invariant.
- No new dependencies.
- Tests reported 602/602 vitest, 13/13 e2e, build green.

**Commit 2 — `4a1d166` app.js + expanded bootstrap preload.** APPROVE.
- `Application` factory now ESM. Vendor reads (`globalThis._`, `globalThis.jQuery`, `globalThis.numeral`) deferred to factory-body time — this matches the new browser load order (vendor `<script>` tags before the bundle IIFE; bundle IIFE before any factory call).
- `app.loadViews()` rewritten: templates are pre-compiled at bundle time via `import.meta.glob('../views/*.html', { query: '?raw', import: 'default', eager: true })`, then bound through underscore's template engine on first `Application` factory call. The legacy `D` variable name is preserved (templates still write `D.foo`), and `_.template(text, null, { variable: 'D' })` correctly passes `null` for data to get a precompiled function back from underscore 1.5.1's signature. `loadViews` keeps a Deferred-returning shape for callers that chain `.then()`.
- Critical commentary in the message about RequireJS's CommonJS-magic factory-body string-scan no longer auto-discovering `require('foo')` is correct, and the explicit additions to `bootstrap.js`'s preload list (Game, Render, underscore, numeral) reflect that.
- Build-test fix-up (`tests/build/build.test.js`) drops the obsolete `dist/scripts/app.js exists` assertion since app.js is now bundled into esm-bundle.js. Correct.

**Commit 3 — `1aeb4a6` Game.js (5703 LOC) ESM port.** APPROVE.
- Wrapper-only conversion as advertised: +32 / -33 lines on a ~5700-line file. The internal `var Game = function() { ... }` factory body is byte-identical; only the `define(function(require) { ... })` shell, the trailing `getGame` glue, and the inline `require('Render').getRender()` / `require('app').getApplication()` calls were touched.
- Two inner `require('i18n')` / `require('Game')` calls inside the reset/reload callback hoisted to module-scope (`i18n` from import, `getGame` as a sibling export). Correct — the deferred-require pattern existed because RequireJS factory string-scan would otherwise eagerly load `Game` at its own module load (causing self-reentry); with ESM that's a non-issue.
- Internal class definitions (`Game.City`, `Game.Project`, ...) keep their class shape — no scope creep into a class refactor.
- No top-level side effects (`Game.init()` remains caller-driven from `app.start()`).
- `import * as bootMod from './boot.js'` preserves the boot-module surface that the Topscores view uses for `subscribePeersChanged`.
- Vendor reads stay at factory-body time (`globalThis._`, `globalThis.jQuery`, `window.sprintf`).

**Commit 4 — `1d57610` Render.js (5305 LOC) ESM port.** APPROVE.
- Wrapper-only conversion: +34 / -25 lines on a ~5300-line file.
- Verified the suspected circular dep was a false alarm: `Render.js` imports only `app`, `setup`, `util` — it does **not** import `Game`. `Game → Render → app` is a one-way edge with `app` constructed first. The PR description's claim matches the diff.
- `globalThis.createjs` aliasing for `Easel` / `Tween` / `Sound` reproduces the legacy AMD shim (which returned `createjs` vs `createjs.Tween` depending on module id). Correct.
- The `Ticker.setFPS` shim and the rest of the factory body are byte-identical.

**Commit 5 — `571c768` retire AMD plumbing (closes #58).** APPROVE.
- All AMD vendor loaders + `require.config.js` deleted from the tree.
- `bootstrap.js` ported to ESM; the `tpl!` plug-in replaced by `import.meta.glob('../views/*.html?raw', { eager: true })` + `_.template(text, null, { variable: 'D' })`.
- `if (typeof window === 'undefined') { /* skip */ }` guard makes the module import-safe in vitest. The `if (typeof webxdc !== 'undefined') boot()` early-fire matches the prior behavior that `esm-entry.js` had been doing — moving the kick-off into `bootstrap.js` keeps the listener-registers-before-UI-runs invariant intact.
- `vite.config.js` drops the `amdBridgeFooter`, `preserveEntrySignatures: 'exports-only'`, and the `amdScripts` static-copy list. `views/` is no longer copied into `dist/` (templates are bundle-inlined). All correct.
- `index.html` now loads vendor libs in dependency order via plain `<script>` tags before the deferred `scripts/esm-bundle.js`. The order matters and is preserved (jquery → jquery-migrate → underscore → numeral → numeral-de → sprintf → easeljs → tweenjs → soundjs → zynga-animate → zynga-scroller).
- `esm-entry.js` adds a tiny `window.require` shim for the existing playwright e2e specs (which inherited an AMD-shaped reach into engine internals). It is ~20 LOC, exposes only `boot` / `LocalEngine` / `app`, and is not a generic AMD loader. This is a pragmatic compatibility shim, not a violation of the "AMD eliminated" goal — it has no module registration, no script fetching, no transitive dep walking. Acceptable; document or convert the e2e specs in a follow-up if you'd like to remove it eventually.
- `tests/build/build.test.js` adds explicit **negative** assertions for the deleted files (`require.config.js`, `bootstrap.js`, `app.js`, `Game.js`, `Render.js`, `vendor/requirejs.js`, `vendor/text.js`, `vendor/tpl.js`, `vendor/jquery-mobile.js`, `vendor/almond.js`, `vendor/native-console.js`) and asserts `index.html` no longer contains `vendor/requirejs.js` or `require.config`, plus the bundle no longer contains `define.amd`. These are exactly the right tests to lock in the AMD elimination — nice work.

---

### Non-blocking notes (no changes requested)

1. **PR body bundle-size diff:** the brief asked the implementer to include a one-line size diff. The PR body is truncated in the API response so I can't tell whether it's there; if not, an edit adding "before / after `du -sk dist/`" would be a useful artifact for #58's history. Total line delta is +595 / -3642 which is consistent with a meaningful shrink.
2. **`window.require` shim in `esm-entry.js`** — file an issue if you want to clean up the e2e specs and remove the shim entirely. Out of scope for #58 and not blocking.
3. **`scripts/util.js wait()`** still uses `setTimeout` and is preserved byte-for-byte. Not a "game cycle" timer (just a generic delay helper), so invariant 1 doesn't apply, but if it's truly unused in the post-AMD codebase, it could go in a separate cleanup pass.

---

Wave 7b ready to ship. Once #156 merges, I'll post the "Wave 7b is complete. AMD eliminated; codebase is single-module-system." comment on #58 — flag me when it's in.

---

experintellia (commented):
**Follow-up review on `d9bebd8` (post-review simplification): verdict APPROVE (posted as comment per usual self-approval block).**

Re-verified the same invariants on the new head — all still green:

- **Engine layer untouched.** `LocalEngine.js` diff in this commit is a 3-line header reword; `state.ts`, `materializer.ts`, `clock.ts` are not in the file list. Invariant 3 holds.
- **AMD-survivor grep on `d9bebd8`** returns zero matches in `scripts/` and `index.html`.
- **`webxdc.sendUpdate` emit sites unchanged** — only `LocalEngine.js:151` (`triggerAchievement`, Wave 7a) and `LocalEngine.js:165` (`_persistDelta` Node fallback). No new emit sites.
- **`addr === state.addr` guard intact** at `state.ts:750`.

### Per-file notes

- **`bootstrap.js`** — wrapper drop is clean: single `if (typeof window !== 'undefined' && typeof document !== 'undefined') { ... }` block replaces the `runBrowserBootstrap()` IIFE-style function. The double-compile of `loader.html` is correctly removed; `app.renderView('loader.html')` works because the existing `import.meta.glob('../views/*.html', ...)` already includes `loader.html`. `getApplication()` is now invoked earlier in the boot path (before the engine-replay promise resolves), but that just runs the `Application` factory which has no dependency on replay state — fine. The change-detection on the progress poll (`p.serial !== lastSerial || p.max_serial !== lastMax`) is a real DOM-write reduction at no cost.

- **`app.js`** — `compileTemplates()` losing the lazy `if (_templates) return _templates` is correct: `getApplication()` is already a memoised singleton so the factory body (and therefore `compileTemplates()`) runs exactly once. Comment trims are appropriate; the kept WHYs (TDZ ordering for `globalThis._`, underscore 1.5.1 signature quirk, the `.done()/.fail()` Deferred wrap rationale) are the right ones to preserve.

- **`esm-entry.js`** — extracted `lookup()` helper, `Object.hasOwn()` swap, and `cb(...resolved)` spread are all behavior-preserving and safer (the `in` operator would have walked `Object.prototype`, so `'toString'` or `'constructor'` would have leaked through). Good catch.

- **`Game.js` / `Render.js` / `vendor-install.js` / both test files** — comment-only.

- **`vite.config.js`** — `bundleEsmDev`'s `await rebuild()` removal makes `listen()` non-blocking. Verified the middleware fallback at `vite.config.js:131-134` returns 503 with `// esm-bundle build pending or failed` if `bundleSrc` is null. **One non-blocking caveat:** `<script defer src="scripts/esm-bundle.js">` in `index.html` does *not* auto-retry on a 503 — if a user hits `localhost:3000` faster than the initial Rollup build finishes (cold-start), the page silently fails to boot until they manually refresh. In practice the human-typing-into-browser delay is usually longer than the build, so it'll work most of the time. If you ever see "first load is blank, refresh works" complaints, the fix is to have the middleware `await buildPromise` when `bundleSrc` is null instead of 503-ing immediately. Not a blocker for this PR.

Wave 7b still ready to ship. Will post the "Wave 7b is complete" comment on #58 once #156 lands.

---

PR #168, opened 2026-05-04
feat(ts): convert boot wiring to strict TypeScript

experintellia:
PR 3 of issue #147.

## Summary

Converts the boot-wiring layer to strict TypeScript:

| File (before) | File (after) | LOC | Notes |
|---|---|---|---|
| `scripts/boot.js` | `scripts/boot.ts` | 177 | Pure ESM, no DOM. |
| `scripts/bootstrap.js` | `scripts/bootstrap.ts` | 71 | jQuery-driven loader UI hand-off. |
| `scripts/esm-entry.js` | `scripts/esm-entry.ts` | 33 | Bundle entry + playwright `window.require` shim. |
| `scripts/app.js` | `scripts/app.ts` | 193 | Application factory, vendor globals. |
| `scripts/setup.js` | `scripts/setup.ts` | 29 | Pulled in alongside app — small dependency. |
| `scripts/setup_local.js` | `scripts/setup_local.ts` | 3 | Trivial. |
| `scripts/i18n.js` | `scripts/i18n.ts` | 71 | Pulled in alongside app — gettext layer. |

Plus a one-line export from `scripts/state.ts` (promote `GameSeed` from internal to exported) so `boot.ts` can type its `defaultGame` parameter without redeclaring a near-identical shape.

## Type coverage highlights

- **`boot.ts`**: `BootOptions` / `ReplayProgress` / `PeersChangedSubscriber`. Listener callback receives `Webxdc<unknown>` updates and narrows via `applyDelta`'s existing runtime guard. Reuses `LocalState` and `GameSeed` from `state.ts`.
- **`bootstrap.ts`**: jQuery surface declared as local `JQueryLike` / `JQueryStatic` interfaces (vendor lib not on a typed import path); `loaderEl` typed as `HTMLProgressElement | null`; `showFatal` narrows `unknown` errors via `in`-checks instead of casting.
- **`esm-entry.ts`**: `TestRequire` type captures the legacy `require([...], cb)` shape used by Playwright specs. Single cast-through-unknown on the `window.require` write to avoid colliding with `@types/node`'s `Require` type.
[...]

PR #171, opened 2026-05-04
feat(ts): extract Set collection helper to scripts/game/OrderedSet.ts

experintellia:
PR 4 of issue #147 — first sub-class extraction from `scripts/Game.js`.

## Why a foundation PR before `Game.<SubClass>` typings

`Game.js` is 6088 LOC in one IIFE with 22 class constructors that share helpers (`extend`, `getById`, `getFirstId`, `getByGestalt`, `mergeData`, …) and a common `GameNode` base. The brief's "split sub-class by sub-class" plan needs an established import pattern before subclass extractions can proceed; this PR establishes that pattern with the smallest, most decoupled helper class — `Set` (a pre-ES2015 ordered collection).

TypeScript has no block-level `@ts-nocheck`, so "remove `// @ts-nocheck` from the converted sub-class scope" works only by moving the sub-class to its own file. PR 5 will do the same for `GameNode` (the foundational base class — needed before any subclass extraction). PR 6+ moves out individual subclasses (`Topscores`/`Topscore` first, per the brief's "smallest classes first" rule).

## What changed

**New file: `scripts/game/OrderedSet.ts` (62 LOC, fully strict-typed).**

`OrderedSet<T>` matches the legacy contract:
- `set: T[]` — public underlying array. Legacy `AniTicker.listeners.set` is reassigned via `_.shuffle(...)` at `scripts/Game.js:329,348`, so the underlying array is swappable from outside. Documented in the class docstring.
- `length: number` mirrors `set.length`, snapshotted at the last `add`/`prepend`/`remove`/`each` call. Matches the legacy "stale-friendly" behaviour where `length` doesn't auto-track external `set` mutations (the AniTicker shuffle pattern relies on this).
- `add` / `prepend` / `remove` / `each(func)` with the original signatures. `each(undefined)` is a no-op (legacy guard preserved).

**`scripts/Game.js`:**
- Add `import { OrderedSet } from './game/OrderedSet.js';` at the top.
- Inside the `var Game = function () { … }` factory: single alias `var Set = OrderedSet;` (with the existing `biome-ignore lint/suspicious/noShadowRestrictedNames` carried over to the alias).
- Remove the 38-line inline `var Set = function(set) { … }` + 4 prototype methods.
- All 6 `new Set()` call sites and the `.set` / `.length` reads keep working unchanged.

## Discipline

- **Zero behavioural change.** Same constructor identity, same prototype methods, same arity, same `length` semantics.
- **`Game.js` keeps `@ts-nocheck`.** The new module is strict-typed on its own; the rest of `Game.js` stays quarantined until later PRs.
- **No `as Foo` casts**, no `!` non-null assertions, no `any`.
- No new `setState` / `webxdc.sendUpdate` sites — pure helper extraction.
- No DOM, no clock, no engine touches.

## Test plan

- [x] `pnpm exec biome check .` — clean (only the gitignored `test-results/.last-run.json` local artifact flags, never present in CI).
- [x] `pnpm test:e2e` — 16 specs pass, 1 skipped, 0 failures (`Set` is on the hot path for `GameRoot`'s `children`, `Database.queue`, `Topscores.queue`, `APTicker.listeners`, `AniTicker.listeners`; the e2e suite exercises all of those).
- [ ] CI green on PR.

## Roadmap visibility

- PR 5 (next): extract `GameNode` base class and the shared helpers it captures (`extend`, `_ids`, `getById`, `add`, `get`, `remove`, `clear`) into `scripts/game/GameNode.ts`. This is the foundational refactor — every subclass extends `GameNode`, so it has to be importable before any subclass can move.
- PR 6+: `Topscores` + `Topscore` (~135 LOC, smallest extends-`GameNode` pair, per the brief's preferred ordering), then `Missions` + `Mission`, then `Imperium`, then `Database`, then individual `Perp` classes.

Refs #147.

---

### DIFF
--- a/scripts/Game.js
+++ b/scripts/Game.js
@@ -5,6 +5,7 @@ import { getRender } from './Render.js';
 // alongside scripts that run before vendor `<script>` tags execute.
 import appModule from './app.js';
 import * as bootMod from './boot.js';
+import { OrderedSet } from './game/OrderedSet.js';
 import i18n from './i18n.js';
 import setup from './setup.js';
 import { getTypeSettings } from './type_settings.js';
@@ -20,6 +21,12 @@ var Game = function () {
   var extend = utilDefault.extend;
   var Render = getRender();
   var typeSettings = getTypeSettings();
+
+  // Legacy in-IIFE alias for the extracted ordered-collection class.
+  // See scripts/game/OrderedSet.ts for the full contract; the rest of
+  // this file calls `new Set()` to retain the historical name.
+  // biome-ignore lint/suspicious/noShadowRestrictedNames: legacy collection class predates ES6 Set
+  var Set = OrderedSet;
 
   //////////////////////////////////////////
   //
@@ -73,49 +80,6 @@ var Game = function () {
     app.game = new GameRoot();
     app.game.loadGame(data);
     return app.game;
-  };
-
-  ////////////////////////////////
-  // The Set
-  ////////////////////////////////
-
-  // biome-ignore lint/suspicious/noShadowRestrictedNames: legacy collection class predates ES6 Set
-  var Set = function (set) {
-    if (!set) {
-      set = [];
-    }
-    this.set = set;
-    this.length = this.set.length;
-    return this;
-  };
-
-  Set.prototype.add = function (node) {
-    this.set.push(node);
-    this.length = this.set.length;
-  };
-
-  Set.prototype.prepend = function (node) {
-    this.set.unshift(node);
-    this.length = this.set.length;
-  };
-
-  Set.prototype.remove = function (node) {
-    var index = this.set.indexOf(node);
-    if (index !== -1) {
-      this.set.splice(index, 1);
-    }
-    this.length = this.set.length;
-  };
-
-  Set.prototype.each = function (func) {
-    if (!func) {
-      return;
-    }
-    for (var n = 0; n < this.set.length; n++) {
-      var node = this.set[n];
-      func(node);
-    }
-    this.length = this.set.length;
   };
 
   //////////////////////////////////////////
--- a/scripts/game/OrderedSet.ts
+++ b/scripts/game/OrderedSet.ts
@@ -5,8 +5,8 @@
 // `length` field, single `each` walker).
 //
 // Extracted from scripts/Game.js's IIFE in the issue #147 / Phase 7
-// migration.  Game.ts re-exports this as `Set` to keep the in-IIFE call
-// sites unchanged while the rest of Game.ts is still under `@ts-nocheck`.
+// migration.  Game.js re-exports this as `Set` to keep the in-IIFE call
+// sites unchanged while the rest of Game.js is still under `@ts-nocheck`.
 //
 // Behaviour preserved from the legacy class:
 //   - `set` and `length` are public.  AniTicker.listeners.set is reassigned
@@ -55,8 +55,7 @@ export class OrderedSet<T> {
 
   each(func: (node: T) => void): void {
     if (!func) return;
-    for (let n = 0; n < this.set.length; n++) {
-      const node = this.set[n] as T;
+    for (const node of this.set) {
       func(node);
     }
     this.length = this.set.length;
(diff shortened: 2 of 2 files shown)


experintellia (commented):

PR 4 of #147 — first sub-class extraction from `Game.js`. Establishes the pattern for the rest of the Game.<Class> migrations. Scoped to one helper class, behaviorally equivalent, fully typed.

### Behavioral equivalence — verified

Legacy `Set` (Game.js:75-127) → `OrderedSet<T>` (`scripts/game/OrderedSet.ts`):

| Aspect | Legacy | New | Match? |
|---|---|---|---|
| Constructor with no arg | `set = []` | `initial ?? []` | ✓ Type signature forbids non-array, so `??` is sufficient. |
| `set` field | Public mutable array | `set: T[]` public | ✓ |
| `length` field | Snapshot at last add/prepend/remove/each | Same | ✓ Critical for the AniTicker `_.shuffle(this.set)` pattern at Game.js:329,348. Documented in docstring. |
| `add` / `prepend` / `remove` | push / unshift / indexOf+splice + length update | Same | ✓ |
| `each(undefined)` | Early return | `if (!func) return` | ✓ Legacy guard preserved. |
| `each` loop | Indexed `for` with `set[n]` | Same with `as T` cast for `noUncheckedIndexedAccess` | Behaviorally same (passes undefined slots to callback). |
| Constructor return | Explicit `return this` | Implicit | ✓ JS constructors return `this` either way. |

The IIFE alias `var Set = OrderedSet` at Game.js:28 means every legacy `new Set(…)` and `set.length` / `set.set` read continues to work without touching the call sites.

### Discipline

- **Cast count: 1** (`this.set[n] as T` in `each`). The alternative under `noUncheckedIndexedAccess` is either `!` (disallowed) or a `if (node !== undefined)` guard that would skip sparse-array slots — which **would change behavior** vs the legacy indexed loop. The cast preserves legacy semantics. Non-blocking suggestion: `for (const node of this.set) func(node);` would type `node` as `T` directly without the cast and would be equivalent for non-sparse arrays. Either way is fine.
- **No `!` non-null assertions, no `: any`, no escape hatches.** ✓
- **No new emit sites, no new `setState` writers, no DOM, no clock, no engine touches.** Pure helper extraction. ✓
- **Game.js stays `@ts-nocheck`** — the new module is strict-typed on its own; quarantine on the rest of Game.js is unchanged. ✓
- **biome-ignore on the `Set` alias** preserved with the legacy reason. ✓

### Architecture invariants

- No `setTimeout` for game cycles. ✓
- State source-of-truth = `webxdc.sendUpdate` history. ✓
- No new `webxdc.sendUpdate({info})` emit sites — diff doesn't touch any emit code. ✓
- `addr === state.addr` guard in `state.ts` not touched. ✓
- No DOM globals in handler bodies — pure data-structure module. ✓
- Pure engine layer untouched. ✓
- `logAction` / `resetGame` policy preserved. ✓

### Roadmap visibility

The PR description's planned ordering matches the brief's "smallest classes first" rule:
- **PR 5**: `GameNode` base class + shared helpers (`extend`, `getById`, `add`, `get`, `remove`, `clear`). Foundational — every subclass extends it, must land before any subclass extraction.
- **PR 6+**: `Topscores` + `Topscore` (smallest extends-`GameNode` pair, ~135 LOC), then `Missions` + `Mission`, then `Imperium`, then `Database`, then individual `Perp` classes.

That's the right shape. Each subclass extraction will be a small, focused PR — reviewable.

### Notes (non-blocking)

- Doc typo in `OrderedSet.ts:9`: "Game.ts re-exports this as `Set`" — Game is still `Game.js`, not `Game.ts` yet. Update to "Game.js re-exports" or "Game.ts (when migrated) re-exports."
- The `as T` cast in `each` at line 56 — could be `for (const node of this.set)` to drop it. Marginal; non-blocking.
- The PR claims `pnpm test:e2e` 16 specs pass — `Set` is on the hot path for `GameRoot.children`, `Database.queue`, `Topscores.queue`, `APTicker.listeners`, `AniTicker.listeners`, so the e2e coverage is meaningful here. Trusting.
- CI status: pending. Want green before merge.

### Coordination

PR #168 (boot wiring) merged. No collisions with this PR. Clean ground for the Game.<Class> extraction series.

---

PR #172, opened 2026-05-04
feat(ts): extract GameNode base class + registry + mergeData to scripts/game/

experintellia:
PR 5 of issue #147 — second sub-class extraction from `scripts/Game.js`. **Foundational**: every `Game.<SubClass>` in the IIFE extends `GameNode`, so it has to be importable before any subclass can leave the IIFE in PR 6+.

## What changed

**New: `scripts/game/GameNode.ts` (~370 LOC, fully strict-typed under `noUncheckedIndexedAccess` + `exactOptionalPropertyTypes`).** Exports:

- The instance registry — `_instances`, `_ids`, plus `add` / `get` / `remove` / `clear` helpers. Mutated by `GameNode.init()` / `GameNode.remove()`. Module-level state shared with `Game.js`'s in-IIFE helpers (`getById`, `getByGestalt`, …) via live ESM bindings. Sparse-friendly: cleared slots stay `undefined` so `_id` indices remain stable, matching legacy.
- The `GameNode` class itself — class syntax this time (was a function constructor + `Set.prototype.X` assignments). Subclasses still extend via the legacy `extend(SubClass, GameNode)` helper from `util.js`; ES classes work fine with the prototype-chain pattern that helper sets up. Subclass-injected hooks (`extendEventHandlers`, `extendRender`, `APTick`, `AniTick`) are declared as optional fields so the base class can call them without per-subclass casts.
- `GameNodeConfig` type. Render-side types (`RenderConfig`, `RenderData`, `RenderNodeLike`, …) are local — when later PRs extract `Render` subclasses these will move/widen.
- Local `_$()` helper to narrow `jQuery ?? globalThis.$` once.

**New: `scripts/game/mergeData.ts` (52 LOC).** Pure data-merge helper used by `GameNode.addType` and ~30 other call sites in `Game.js`. Preserves legacy semantics: shallow-merge with `instance_data` winning, plus the Project-like deep-clone-tokens-and-overwrite-amounts branch.

**`scripts/Game.js`:**
- New imports for `GameNode`, registry helpers, `mergeData`.
- Removed inline `var GameNode = function(config)` plus all 25 prototype methods that I moved (lines 296-578).
- Removed inline `var mergeData = function(…)` plus the registry (`var _instances`, `_ids`, `add`, `get`, `remove`, `clear`).
- Subclass-side prototype additions (`GameNode.prototype.openGenericPopup` at line 1484, `Error/NoCash/NoAP` mixins later in the file) keep working — they attach to the imported class via live-binding mutation, the same way they did inside the IIFE.

**`types/env.d.ts`:**
- `JQueryStatic` call signature widened to include `object` so `$(this)` (event-bus wrapping for `GameNode` instances) type-checks.
- `JQueryLike` gains `off(ev?: string)` so the `GameNode.off()` method type-checks.

- No new `setState` writers — `boot.ts`'s listener callback is still the sole production `setState`; `GameNode.init()` doesn't touch state.
- No DOM globals in handler bodies — this is the GameNode UI layer, the engine layer (`LocalEngine.ts`) is untouched.
- `addr === state.addr` guard in `state.ts` unchanged.
- `logAction` policy preserved, `resetGame` still removed.

## Discipline

- Zero internal `as Foo` casts in either new file. The few `as` uses are at narrow boundaries: `getType()`'s loose return shape, the `this.GameRoot` cast for forward-ref to GameRoot's wider interface (which lives in `Game.js`), and one `Object.create(null)` widening.
- Zero `!` non-null assertions outside the legitimate definite-assignment field markers (`_id!`, `id!`, etc. — class-init pattern, not value assertions).
- Zero `: any`.
- All error-handling identical to the legacy methods — no new defensive guards, no behavioural drift.

## Test plan

- [x] `pnpm test:e2e` — 16 specs pass (exercises `GameNode.init`, `remove`, `addType`, `makeRenderConfig`, `render`, full event-bus surface via topscores/share-merge/sweepstakes/zoom-controls specs).
- [ ] CI green on PR.

## Pattern established for PR 6+

- New `scripts/game/<SubClass>.ts` imports `GameNode` + the legacy `extend` helper from `util.js`.
- Module-level `extend(SubClass, GameNode)` sets up the prototype chain just as inside the legacy IIFE.
- `Game.js` drops the inline subclass definition and adds the import.
- Each subclass extraction is small (50-200 LOC) and reviewable.
- PR 6: `Topscores` + `Topscore` (smallest extends-`GameNode` pair, ~135 LOC).

Refs #147.

---

### DIFF
--- a/scripts/Game.js
+++ b/scripts/Game.js
@@ -5,7 +5,9 @@ import { getRender } from './Render.js';
 // alongside scripts that run before vendor `<script>` tags execute.
 import appModule from './app.js';
 import * as bootMod from './boot.js';
+import { GameNode, _ids, _instances, add, clear, get, remove } from './game/GameNode.js';
 import { OrderedSet } from './game/OrderedSet.js';
+import { mergeData } from './game/mergeData.js';
 import i18n from './i18n.js';
 import setup from './setup.js';
 import { getTypeSettings } from './type_settings.js';
@@ -43,37 +45,11 @@ var Game = function () {
   /////////////////////////////////////////////
   // Some generic tools and getter functions
   /////////////////////////////////////////////
-
-  var _instances = [];
-
-  var _ids = {};
-
-  var add = function (node) {
-    _instances[node._id] = node;
-    _ids[node.id] = node;
-  };
-
-  var get = function (_id) {
-    return _instances[_id];
-  };
-
-  var remove = function (_id) {
-    if (get(_id)) {
-      delete _ids[get(_id).id];
-    }
-    _instances[_id] = undefined;
-  };
-
-  var clear = function () {
-    // Clear everything that has been instantiated so far
-    for (var n = 0; n < _instances.length; n++) {
-      var node = _instances[n];
-      if (node) {
-        node.remove();
-      }
-    }
-    _instances.length = 0;
-  };
+  //
+  // _instances / _ids registry + add/get/remove/clear were extracted to
+  // scripts/game/GameNode.ts so the GameNode base class can mutate them
+  // without an IIFE-closure round-trip.  The imports above bring them in
+  // as live bindings; the in-IIFE call sites below keep the legacy names.
 
   var init = function (data) {
     // Inits GameRoot as a Singleton, for now.
@@ -152,35 +128,7 @@ var Game = function () {
     return full_type.split(setup.typeSeparator)[1];
   };
 
-  var mergeData = function (type_data, instance_data) {
-    // Merges data objects, second argument overwrites the first ones and returns new merged object
-    var data = {};
-    _.extend(data, type_data);
-    _.extend(data, instance_data);
-    // If the type data is Project-like (has powerups and tokens) merge token amounts
-    if (
-      instance_data &&
-      type_data &&
-      instance_data.hasOwnProperty('powerups') &&
-      type_data.hasOwnProperty('tokens') &&
-      instance_data.hasOwnProperty('tokens')
-    ) {
-      var tokens = [];
-      // No way around this: make a deep copy to avoid messing with token type_data
-      _.each(type_data.tokens, function (v, k) {
-        tokens[k] = _.clone(v);
-      });
-      var instokens = instance_data.tokens;
-      _.each(tokens, function (t, k) {
-        var overwrite = _.findWhere(instokens, { gestalt: t.gestalt });
-        if (overwrite && overwrite.amount) {
-          t.amount = overwrite.amount;
-        }
-      });
-      data.tokens = tokens;
-    }
-    return data;
-  };
+  // mergeData was extracted to scripts/game/mergeData.ts (imported above).
 
   var convertPowerupType = function (game_type) {
     // This costs a beer dudes...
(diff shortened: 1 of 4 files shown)


experintellia (commented):

Foundational PR 5 of #147 — `GameNode` base class + registry + `mergeData` extracted from `Game.js`'s IIFE. The shape is right and architecture invariants are preserved. Cast count is on the high side at 16, but most are anchored on future extractions and should dissolve in PR 6+; calling that out in the notes.

### Behavior preservation

- **`OrderedSet`-style snapshot semantics, prototype mixin pattern**: legacy `GameNode.prototype.X = ...` mutations at Game.js:1199 (`openGenericPopup`), 2683 (`initPopupEvents`), 3767 (`fetchProvided`), 4810 (`Error`), 4824 (`NoCash`), 4832 (`NoAP`) attach to the imported class via ESM live bindings. Confirmed the imports in Game.js + the prototype-property writes will mutate the imported class's prototype object; subsequent imports of the class see the augmented prototype. Works, but it's a code smell — flagging below as a non-blocking cleanup target for later PRs.
- **`extend(SubClass, GameNode)` legacy helper from `util.js`** continues to work with the new ES6 class. ES6 class prototype chains are interoperable with the function-constructor pattern that `extend` sets up. Subclasses extending `GameNode` won't see semantic changes.
- **`mergeData.ts`** preserves the two-branch behavior exactly: shallow merge with `instance_data` winning, plus the Project-like deep-clone-tokens-and-overwrite-amounts branch. Behavior identical to legacy at Game.js:155-183.
- **One subtle ES6 class vs function-constructor difference**: ES6 class methods are non-enumerable on the prototype, while function-constructor prototype assignments are enumerable. If any code did `for (key in instance)` or `Object.keys(GameNode.prototype)`, it would see different keys. The PR claims `pnpm test:e2e` passes 16 specs that exercise GameNode heavily — trusting that signal. Worth keeping in mind when future PRs extract subclasses.

- **No new `webxdc.sendUpdate` emit sites.** Zero in both new files. ✓
- **No new engine `setState` writers.** The `setState(state: string, value: boolean): void` method at GameNode.ts:286 is the **UI-state setter** for per-instance flags ("selected", "highlighted", etc.), not the engine `boot.setState`. Same naming as the legacy method; not a new writer of the engine state. ✓
- **No new `setTimeout` for game cycles.** ✓
- **No DOM globals in handler bodies** — `LocalEngine.ts` untouched. This is the GameNode UI layer, separate from the handler layer. ✓
- **Pure engine layer untouched.** `state.ts`, `materializer.ts`, `clock.ts`, `LocalEngine.ts`, `boot.ts` all unchanged. ✓

### Cast inventory — 16 in `GameNode.ts`, 0 in `mergeData.ts`

Categorized:

| Pattern | Count | Verdict |
|---|---:|---|
| `this.GameRoot as GameNode & { ... }` (forward-ref to wider GameRoot interface) | 3 | Transitional; will dissolve when GameRoot is extracted. |
| `as Record<string, unknown>` (instance_data / type_data / config dynamic shape) | 7 | Should resolve when `instance_data`, `type_data`, `RenderConfig` get proper interfaces extracted. |
| `getRender() as unknown as RenderModule` + `Render as Record<string, unknown>` | 2 | Will dissolve when `Render.js` is typed. |
| `this as unknown as Record<string, unknown>` in `setAttrs` (line 314) | 1 | Genuinely required — legacy `setAttrs` does dynamic property writes onto the instance. The alternative is an index signature on GameNode itself, which defeats the typing. Justifiable seam. |
| `cfg as RenderConfig` (lines 425-426) — typed-builder finalization | 2 | Acceptable — cfg is built up as `Record<string, unknown>` then narrowed at output. |
| `getTypeSettings() as Record<...>` (boundary on `@ts-nocheck` type_settings.js) | 1 | Boundary cast. ✓ |

The cluster at lines 455-456 (`getRender()` + dynamic `Render[type]` lookup) could be cleaner with a typed `Render.getCtor(name)` helper — non-blocking, but consider adding it to `RenderModule` when `Render.js` is typed.

The brief's "one or two well-justified casts" rule is strained, but each cast here is anchored on a future extraction. I'd want PR 6+ to actively reduce this count — when `GameRoot` is extracted, the 3 forward-ref casts must go; when instance_data/type_data interfaces are formalized, the 7 `Record<string, unknown>` reads must go. **Track this on the WW thread roadmap** so the count goes down, not up, as extractions accumulate.

### `!:` field markers — acceptable

`_id!`, `id!`, `children!`, `states!`, `GameRoot!`, `jq!` are class-field **definite-assignment markers**, not value assertions. The brief's "no `!` non-null assertions" was about value assertions (`x!.foo`) that mask nullability. Field markers say "this field is initialized in `init()` even though TS can't see it through that flow" — runtime behavior is identical to declaring without the marker.

Verify these are all set in `init()` (or its overrides) before any read:
- `_id`, `id` — set in `add()` (registry).
- `children`, `states`, `jq` — set in `init()`.
- `GameRoot` — set in `init()` via `(get(0) as GameNode | undefined) ?? this`.

Looks fine.

### Other notes (non-blocking)

- **Doc inconsistency**: `OrderedSet.ts:9` mentions "Game.ts re-exports this as `Set`" but Game is still `Game.js`. Same minor typo from PR #171 — not a bug, just stale wording. (Not in this PR's diff; carry over from before.)
- **Prototype mixin smell**: 6 `GameNode.prototype.X = function (...) {...}` assignments still live in `Game.js` and mutate the imported class. Works, but it's a smell that should be cleaned up as those methods get moved into `GameNode.ts` (or split into mixin modules) on later PRs. Track on the WW roadmap.
- **`types/env.d.ts` widenings**: `JQueryLike.off(ev?: string)` and `JQueryStatic`'s call signature accepting `object` are both well-justified by the legacy `$(this)` event-bus pattern on GameNode. Reasonable widenings. ✓
- CI status: pending. Want green before merge.

### Coordination

PR #171 (`OrderedSet`) merged. No collisions with this PR. Next up per the implementer's roadmap: PR 6 with `Topscores` + `Topscore` (~135 LOC) — smallest extends-GameNode pair. Sensible ordering.

Approved for merge once CI is green. **Heads up for PR 6+**: each extraction must reduce the GameNode.ts cast count, not just add more in the new file. The forward-ref / `Record<string, unknown>` patterns should converge to typed shapes as their owning classes get extracted.

---

PR #173, opened 2026-05-04
feat(ts): extract Topscores + Topscore subclasses + id-helpers

experintellia:
PR 6 of issue #147 — first GameNode subclass extraction, plus the read-side id/path/gestalt helpers that subclasses need.

## What changed

**New: `scripts/game/Topscores.ts` (135 LOC, fully strict-typed).** ES class `extends GameNode`. Replaces the legacy `var Topscores = function … + extend(Topscores, GameNode)` pattern; ES `extends` produces an equivalent prototype chain. `initTopscore()` / `updateScores()` / `extendRender()` / `extendEventHandlers()` preserved bit-for-bit. Live leaderboard refresh via `bootMod.subscribePeersChanged()` unchanged. **Eliminated the circular `Game.Topscore` lookup** in favour of a direct `import { Topscore } from './Topscore.js'`.

**New: `scripts/game/Topscore.ts` (115 LOC, fully strict-typed).** ES class `extends GameNode`. `fetchScore()` preserves the 30-second cache, the Deferred `.done()/.fail()` chain (`LocalEngine.getRanking` still returns a jQuery Deferred via `app.remote` wrapping), and the `user_in_top` derivation. Uses `appModule.getApplication()` lazily inside `fetchScore` to avoid the `app.js → Game.js → Topscore.ts` module-load cycle (caught the cycle when `tests/toolchain.test.js` failed; defer-to-call-time fixed it).

**`scripts/game/GameNode.ts` (extracted in PR #172) — gains 12 read-side helpers** that `Game.js`'s IIFE used to inline:
- `getById`, `getFirstId`, `getLastId`, `getParentId`, `getByFirstId`, `getByLastId`, `getParentFromPath`, `getGestalt` (path/id helpers)
- `getByGestalt`, `getAllByGestalt`, `eachByGestalt`, `getByType` (gestalt/type queries)

Each reads from the registry's `_ids` exported alongside `GameNode` itself; live bindings mean `Game.js`'s other IIFE consumers keep working unchanged.

Also in `GameNode.ts`: `extendEventHandlers` / `extendRender` / `APTick` / `AniTick` are now declared as **optional method members** (`extendEventHandlers?(): void`) rather than function-typed properties so subclasses can `override` them with class-body methods (TS treats those as method-vs-property mismatches under `exactOptionalPropertyTypes` otherwise).

**`scripts/Game.js`:**
- Imports the new `Topscores` / `Topscore` classes + the 12 id-helper functions from `./game/GameNode.js`.
[...]

PR #174, opened 2026-05-04
feat(ts): extract Missions + Mission subclasses to scripts/game/

experintellia:
PR 7 of issue #147 — second `GameNode` subclass extraction. Same shape as PR #173 (`Topscores`/`Topscore`): two new strict-typed files, one ES class each, the legacy `Game.Mission` lookup replaced by a direct import.

## What changed

**New: `scripts/game/Mission.ts` (~220 LOC).** ES class `extends GameNode`. `getBranch` / `getNext` / `updateRender` / `updateGoal` / `openMissionPopup` / `checkTutorial` / `extendEventHandlers` / `extendRender` preserved bit-for-bit, including the legacy assignment-as-condition `goal.mission = this.gestalt` (split into a guarded `if` to satisfy `exactOptionalPropertyTypes`; runtime semantics identical). `GameRoot` is narrowed via a local `GameRootWithMissions` interface declaring only the fields/methods `Mission` touches (`IPerps`, `raw_data`, `DBTokensAbsolute`, `setState`, `makeNotifications`, `openGenericPopup`, `getOriginGestaltFromOriginTokenGestalt`).

**New: `scripts/game/Missions.ts` (~280 LOC).** ES class `extends GameNode` with **`declare children: OrderedSet<Mission>`** (per the PR #173 reviewer's typed-children pattern) — child reads don't need per-call casts. `initMissions` / `getMission` / `getActive/Visible/Completed/NextMissions` / `checkProjectGoals` / `updateMissions` / `updateMissionGoals` preserved. `app.remote.recheckMissions()` is resolved lazily inside `extendEventHandlers` to dodge the `app.js → Game.js → Missions.ts` module-load cycle (same pattern as `Topscore.ts` in PR #173).

**`scripts/Game.js`:**
- Imports `Mission` and `Missions` from `./game/`.
- Removes ~410 LOC of inline `Missions` + `Mission` class definitions (lines 2934-3343 in the pre-PR file).
- Replaces with a comment block pointing to the extracted files.
- API publisher (`Missions: Missions, Mission: Mission`) at the bottom of the IIFE keeps re-exposing the live class identities; `GameRoot.loadGame`'s `new Game.Missions(...)` call site etc. work unchanged via ESM live binding.

- No new `setState` writers — `Missions`/`Mission` are UI nodes; engine layer untouched.
- `logAction` / `resetGame` still removed.

## Cast count — direction

Per the PR #172 reviewer's heads-up ("each extraction must reduce the cast count, not just add more"):
- **`scripts/game/GameNode.ts`: 18 → 18.** No helpers added; held flat.
- **`scripts/game/Mission.ts`: 8 internal casts.** 6 are `this.GameRoot as unknown as GameRootWithMissions` forward-refs (will dissolve when `GameRoot` is extracted). 2 are `stopPropagation` event-arg narrows (carry-over from the legacy DOM-event surface).
- **`scripts/game/Missions.ts`: 7 internal casts.** 4 are GameRoot forward-refs. 2 are `stopPropagation` event narrows. 1 is the Deferred chain `as unknown as DoneFailChain<RecheckMissionsResult>` for the `recheckMissions` response — boundary on `app.remote`'s loose return shape. Used `declare children: OrderedSet<Mission>` to drop the 3 `child as Mission` casts that would otherwise appear in the children-iteration paths (per the PR #173 reviewer suggestion).

Render-side narrows are minimal here (Mission.ts has 1; Missions.ts has 0 in the new code) because `Mission`'s render surface is just `.render()` + the `FXMissionGoalComplete` fx hook on `GameRoot.renderNode`. Will dissolve when `Render.js` is typed.

## Test plan

- [x] `pnpm test:e2e` — 16 specs pass. `gameplay.spec.ts` and `sweepstakes-upgrade.spec.ts` exercise the `buy_perp` / `collect_profiles` mission-goal flows that go through `Missions.updateMissions` and `Mission.updateGoal`.
- [ ] CI green on PR.

## Roadmap

- PR 8 (next): `Imperium` (~29 LOC, the smallest extends-`GameNode` subclass).
- PR 9: `Database` (~500 LOC).
- PR 10+: Perp classes (`GamePerp` first, then `DatabasePerp`, `CityPerp`, `AgentPerp`, `ContactPerp`, `PusherPerp`, `ClientPerp`, `ProxyPerp`, `ProjectPerp`, `TokenPerp`, `SupertokenPerp`, `ProfileSet`).

Refs #147.

---


experintellia (commented):

PR 7 of #147 — `Missions` + `Mission` extraction. Pattern is consistent with PR #173, and the `declare children: OrderedSet<Mission>` pattern from the prior heads-up was correctly adopted. Architecture invariants preserved. One behavior nuance worth flagging on the legacy assignment-as-condition translation; not blocking but worth being aware of.

### Cast count — direction

| File | Before | After | Change |
|---|---:|---:|---|
| `GameNode.ts` | 18 | 18 | **Held flat.** ✓ |
| `Mission.ts` | (new) | 8 | 4 GameRoot forward-refs, 2 `parentNode` narrows, 1 RenderNode narrow, 1 prototype-mutation cast (see notes). |
| `Missions.ts` | (new) | 8 | 5 GameRoot forward-refs, 1 Deferred chain, 1 fallback empty-object, 1 `Parameters<>` narrow. |

Implementer's PR description estimated 8 + 7 = 15; actual is 8 + 8 = 16. Off by one (one extra GameRoot forward-ref in `Missions.ts`). Same magnitude.

`declare children: OrderedSet<Mission>` at `Missions.ts:53` ✓ — adopted per the PR #173 heads-up. The would-be 3 `child as Mission` child-iteration casts are eliminated.

### Behavior preservation — one nuance

**The `goal.mission = this.gestalt` assignment-as-condition.** Mission.ts:127:
```ts
if (this.gestalt !== undefined) goal.mission = this.gestalt;
if (!goal.mission) return;
```

Legacy (assumed): `if ((goal.mission = this.gestalt)) { … }` — assigns the gestalt to goal.mission unconditionally and uses the truthy assigned value as the guard.

These diverge on the `this.gestalt === undefined` case:
- **Legacy**: `goal.mission = undefined` (assignment happens), then if-condition is falsy, body skipped.
- **New**: assignment skipped (goal.mission keeps prior value), then `if (!goal.mission)` checks the **prior** value — could be truthy if a previous mission set it.

In practice, `this.gestalt` is always defined at runtime (set via the `addType` constructor flow), so the divergence is unreachable. But the inline comment claiming "bit-for-bit" is technically inaccurate. **Strictly bit-for-bit** would be:
```ts
goal.mission = this.gestalt;  // always assign, even undefined
if (!goal.mission) return;
```
…but that may trip `exactOptionalPropertyTypes` if the `goal.mission` field type doesn't include `undefined`.

Non-blocking. The defensive translation is safer than a literal port and the tests pass. Worth either updating the comment to "behaviorally identical when `this.gestalt` is defined (always true at runtime)" OR simplifying to the unconditional assign with a `// @ts-expect-error` if exactOptionalPropertyTypes fights it.

- **No `!` value assertions** in either new file. ✓
- **No `: any`.** ✓
- **No new `webxdc.sendUpdate` emit sites.** ✓
- **No new engine `setState` writers** — Missions/Mission are UI nodes; `boot.setState` not called from either. ✓
- **No new `setTimeout`** for game cycles. ✓
- **No DOM globals in handler bodies.** Handler layer (`LocalEngine.ts`) untouched. ✓

### Notes (non-blocking)

**1. Prototype-mutation cast at `Mission.ts:216`.** This pattern:
```ts
(Mission.prototype as unknown as { popupTemplate: string }).popupTemplate = 'popup_mission.html';
```
preserves the legacy `Mission.prototype.popupTemplate = '...'` from outside the class. It's the same prototype-mutation smell I flagged on PR #172. Easy fix here: move it into the class body as a class field:
```ts
class Mission extends GameNode {
  popupTemplate = 'popup_mission.html';
  ...
}
```
Drops one cast and removes the prototype-mutation. The legacy `Game.js`-side mixins (`openGenericPopup`, `Error`/`NoCash`/`NoAP`) can't move yet because they live in IIFE scope, but the static template-name field is trivially in-class.

**2. Lazy `appModule.getApplication()` to dodge module cycle.** Same pattern as `Topscore.ts` in PR #173 — defer-to-call-time inside `extendEventHandlers`. Pragmatic. ✓

**3. `({} as Record<string, never>)` fallback at `Missions.ts:95`.** This is the empty-object fallback for `getMission(gestalt)` when the lookup fails. Could be tightened to return `Record<string, MissionConfigField | never>` or split the function signature into `getMission(): MissionConfig | undefined` so callers get a real null result instead of an empty-object sentinel. Borderline — flagging because empty-object-as-not-found is a known footgun pattern.

**4. `goal as Parameters<Mission['updateGoal']>[0]` at `Missions.ts:266`.** Smart use of `Parameters<>` — but it's still a cast. The cleaner fix is to type `goal` correctly at the source iteration site so the cast becomes unnecessary. Non-blocking.

### Roadmap

- PR 8: `Imperium` (~29 LOC, smallest). Easy win. Should drop several GameNode.ts forward-refs depending on what `Imperium` exposes.
- PR 9: `Database` (~500 LOC). Big.
- PR 10+: Perp classes.

**Heads up for PR 9 (Database)**: it's the first ~500-LOC subclass and likely to introduce more `as Record<string, unknown>` patterns for instance_data shapes. Worth declaring `DatabaseInstanceData` interface upfront if the field set is enumerable.

### CI

Pending. Trusting `pnpm test` (622), `pnpm test:e2e` (16, incl. `gameplay.spec.ts` + `sweepstakes-upgrade.spec.ts` exercising the mission-goal flows). Want CI green before merge.

---

PR #176, opened 2026-05-04
feat(ts): extract ProfileSet subclass to scripts/game/

experintellia:
PR 9 of issue #147 — fourth `GameNode` subclass extraction.

**Reordering note**: this was originally planned as the `Database` extraction, but `Database` imports `ProfileSet` directly, so landing `ProfileSet` first removes one of `Database`'s cross-class refs and makes PR 10's diff smaller.

## What changed

**New: `scripts/game/ProfileSet.ts` (~250 LOC, fully strict-typed).** ES class `extends GameNode`. Used as both a template for buyToken popups and as the cued object in the `Database` queue (`Database.cue` constructs one and prepends to the queue `OrderedSet`).

Class-side surface preserved bit-for-bit:
- **`ProfileSetConfig`** interface gathers all the constructor-time stamps (`psid`, `origin`, `profiles_value`, `markNew`, `sortByGestalt`, `convertTokens`, `filter_is_query`, `DBAmounts`, `lockAmountZero`, `lockNotInDB`, `markUpgradeValues`, `lastUpgradeValues`) so callers (`Database.cue`, the various Perp `BuyToken` flows) get a typed config object instead of `Record<string, unknown>`.
- `tokens_map`: `Record<string, { amount?, [key]: unknown }>` typed field; `tokens_set: TokenSetEntry[]` typed field with the legacy `diff/done/database/locked/new` fields.
- `GameRoot` narrowed via local `GameRootForProfileSet` interface declaring `DBTokens`, `DBTokensAbsolute`, `profiles_value`, `raw_data`, `getTypeData` — the only fields ProfileSet reads.
- `updateNewMarker` preserved.

Behaviour preservation notes:
- Legacy `tokens = _.clone(tokens)` shallow-clone preserved via `Array.isArray`-branched `Array.map` / object-spread.
- Legacy `0 === 0` always-true sort guard at line 1933 preserved with the same biome-ignore comment from `Game.js` (removal is a separate refactor — the gate-effect of the full condition isn't documented).
- The two non-array branches (token_map from queue/collect vs flat list from the source perp) preserve the original control flow.

**`scripts/Game.js`:**
- Imports `ProfileSet` from `./game/ProfileSet.js`.
- Removes ~145 LOC of inline `ProfileSet` class definition (lines 1830-1974 in the pre-PR file).
- Replaces with a comment block pointing to the extracted file.
- The legacy IIFE never published `ProfileSet` on the API publisher; the inline `new ProfileSet(...)` call sites in `Database.cue` and the Perp `BuyToken` methods still work because `ProfileSet` is now imported at module top — same closure visibility, different resolution path.

- No new `setState` writers — `ProfileSet` is a UI queue entry; engine layer untouched.

## Cast count — direction

- **`scripts/game/ProfileSet.ts`: 5 internal casts.** 3 are `GameRoot` forward-refs (`as unknown as GameRootForProfileSet`). 1 is the `(groot.getTypeData('ProfileSet') || {}) as Record<string, unknown>` fallback for `data`. 1 is `t as unknown as string` for the `convertTokens` compat branch (legacy code overloaded the gestalt slot to hold a string in that branch).

All 5 will dissolve when `GameRoot` is extracted with a typed surface and when `convertTokens` callers tighten their token shape.

Render-side narrows are zero in this file because `ProfileSet` has no Render code (popup template is set up by `Database`, which will keep its Render touches when extracted in PR 10).

## Test plan

- [x] `pnpm exec biome check .` — clean (auto-fix replaced the legacy `var gnode = this` aliases with direct `this` references).
- [x] `pnpm test:e2e` — 16 specs pass. `share-merge.spec.ts` and `gameplay.spec.ts` exercise the integrate-collect flow that walks through `Database.cue → new ProfileSet → mergeCued`.
- [ ] CI green on PR.

## Roadmap update

- **PR 10: `Database` (~500 LOC).** Now imports `ProfileSet` directly. Still has the dynamic `Game[node.game_type]` lookup + a direct `Game.TokenPerp` ref; will solve via a small `perpRegistry` helper module (register/lookup pattern) so `Database` doesn't need to wait for every Perp class extraction first.
- **PR 11+: Perp classes** (`GamePerp` first, then individual perps).

Refs #147.

---

### DIFF
--- a/scripts/Game.js
+++ b/scripts/Game.js
@@ -30,6 +30,7 @@ import { Imperium } from './game/Imperium.js';
 import { Mission } from './game/Mission.js';
 import { Missions } from './game/Missions.js';
 import { OrderedSet } from './game/OrderedSet.js';
+import { ProfileSet } from './game/ProfileSet.js';
 import { Topscore } from './game/Topscore.js';
 import { Topscores } from './game/Topscores.js';
 import { mergeData } from './game/mergeData.js';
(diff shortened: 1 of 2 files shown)


experintellia (commented):

PR 9 of #147 — `ProfileSet` extraction. Smart reordering decision (landing this before `Database` removes one cross-class ref). Architecture invariants preserved. One subtle behavior nuance worth flagging on the legacy `_.clone(tokens)` translation.

### Cast count

| File | Before | After | Change |
|---|---:|---:|---|
| `GameNode.ts` | 18 | 18 | **Held flat.** ✓ |
| `ProfileSet.ts` | (new) | 4 | 2 GameRoot forward-refs, 1 typeData fallback narrow, 1 `t as unknown as string` for `convertTokens` compat. |

PR description estimated 5; my grep counts 4. Either way, all 4 are anchored on future GameRoot extraction or the documented convertTokens-overloaded-string seam.

### Behavior preservation — one nuance

**The `_.clone(tokens)` translation.** Legacy:
```js
tokens = _.clone(tokens);
```
underscore 1.5.x `_.clone` is:
- For an Array: `slice()` — shallow array copy (elements shared by reference).
- For a plain object: shallow `{...obj}` (top-level fields shared).

New code:
```ts
if (Array.isArray(tokens)) {
  workTokens = tokens.map((t) => ({ ...t }));        // ← clones each element TOO
} else {
  workTokens = { tokens_map: { ...tokens.tokens_map } };  // ← spreads inner map TOO
}
```

This is **deeper than `_.clone`**: the array branch clones each element, and the object branch spreads `tokens.tokens_map`. Legacy behavior shared element / inner-map references; new behavior doesn't.

Where this could matter: if any downstream consumer of `workTokens` (e.g. mutations during the `addtokens.forEach` walk) relies on the mutation propagating back to the original `tokens` reference, that propagation no longer happens.

**In practice it's likely harmless** — the immediate downstream step is `addtokens.forEach((token) => { const t: TokenSetEntry = { gestalt: token.gestalt || '', ...token }; ... })`, which creates fresh `TokenSetEntry` objects anyway. So the over-cloning at the `workTokens` initialization is washed out by the time data lands in `tokens_set`. e2e tests pass (`share-merge.spec.ts` + `gameplay.spec.ts` exercise the integrate-collect → `Database.cue` → `new ProfileSet` → `mergeCued` flow).

But the PR description characterizes this as "shallow-clone preserved" — strictly it's not. Either:
- Update the comment to "shallow-clone, deepened by one level for safety," or
- Match legacy exactly: `workTokens = Array.isArray(tokens) ? tokens.slice() : { tokens_map: tokens.tokens_map };` (with appropriate type assertions).

Non-blocking. Worth a note.

- **No `!` value assertions, no `: any`, no new `webxdc.sendUpdate` emit sites, no new engine `setState` writers, no new `setTimeout`, no DOM in handler bodies.** ✓
- **No prototype-mutation smell** — `ProfileSet` is non-rendered (no `renderType` to set), so the cleanup from PR #175's fixup doesn't apply here.

### Other notes (non-blocking)

- **Sort guard `0 === 0`** preserved with the same biome-ignore from legacy. ✓ Right call to keep — removal needs separate analysis of the gate-effect of the full condition.
- **`new ProfileSet(...)` call sites in Game.js** at lines 2052, 3271, 3627, 3631, 3875, 4712 — all resolve to the module-top imported `ProfileSet` directly. The legacy IIFE never published `ProfileSet` on the API, so there's no `Game.ProfileSet` to maintain. Clean. ✓
- **`var gnode = this` aliases removed by biome auto-fix** — sensible cleanup, semantically equivalent.

### Roadmap

The PR 10 `Database` plan with a `perpRegistry` helper module to dodge the dynamic `Game[node.game_type]` lookup is a good architecture call — decouples Database extraction from waiting on every Perp class. Worth a single-purpose register/lookup helper rather than typing `Game[node.game_type]` directly.

### CI

Pending. Trusting `pnpm test` + `pnpm test:e2e` claims. Want green before merge.

---

PR #188, opened 2026-05-05
phase 7 (#147): extract remaining 10 Perp subclasses from Game.js IIFE

experintellia:
## Summary

Follow-up to PR #178 (GamePerp). Extracts the **remaining 10 Perp subclasses** from `scripts/Game.js`'s IIFE into typed `scripts/game/*.ts` modules. Game.js is shrunk by ~1,960 lines net.

PR #178 merged with only the GamePerp shared base; this PR completes the perp hierarchy extraction.

## Stack (PRs 12–16 of issue #147)

| PR | Class(es) | New file(s) |
|---|---|---|
| 12 | `DatabasePerp` + `CityPerp` | `scripts/game/{DatabasePerp,CityPerp}.ts` |
| 13 | `AgentPerp` + `ProxyPerp` | `scripts/game/{AgentPerp,ProxyPerp}.ts` |
| 14 | `ContactPerp` + `PusherPerp` | `scripts/game/{ContactPerp,PusherPerp}.ts` |
| 15 | `ClientPerp` + `ProjectPerp` + `powerupTypes.ts` helper | `scripts/game/{ClientPerp,ProjectPerp,powerupTypes}.ts` |
| 16 | `TokenPerp` + `SupertokenPerp` | `scripts/game/{TokenPerp,SupertokenPerp}.ts` |

After this PR, the `perpRegistry` seam (`setPerpClasses` / `lookupPerpClass`) can retire — direct imports replace the dynamic `Game[node.game_type]` lookup.

## Patterns applied uniformly across all 10 subclasses

- **`protected get groot()` accessor** returning a per-subclass narrow that `extends GameRootForPerp`. Drops 4+ inline casts per subclass.
- **Shared types imported from `GamePerp.ts`**: `RenderNodeLike`, `RenderPopupLike`, `BuyPerpResult`, `ChargeResult`, `DoneFailChain`, `ProvidedPerpRow`, `GameRootForPerp`. Subclasses re-use instead of redeclaring.
- **`protected static GamePerp._stopProp(e)`** for stopPropagation guards (replaces hand-rolled try/typeof checks).
- **`protected GamePerp._serverError(data)`** helper for the 16+ literal `gnode.Error?.('The computer says NOOOO', data)` sites.
- **Class-field overrides** (per PR #175 review pattern): `override renderType = 'Perp'`, `override popupTemplate = '...'`, `override cableType = 'inout' as const`. No prototype assignment.
[...]

PR #193, opened 2026-05-05
phase 7 (#147): retire perpRegistry from Database / DatabasePerp / Game.js

experintellia:
## Summary

PR 17 of issue #147. Retires the `setPerpClasses(Game)` IIFE-end registration call in `Game.js` and migrates Database / DatabasePerp from the dynamic registry to typed direct-map / direct-import lookups.

The seam doesn't fully retire — `GamePerp.BuyPerp` still calls `lookupPerpClass(name)` because GamePerp.ts can't directly import its own subclasses (each one extends GamePerp; direct imports would be a load-time cycle). The registry stays alive only for that one caller, seeded transparently by the new `perpCtors` module.

## What changes

**New module:** `scripts/game/perpCtors.ts` (~50 LOC)
- Imports all 10 perp classes directly.
- Exports a `Record<string, PerpCtor>` keyed by `gameType`.
- Calls `setPerpClasses(perpCtors)` as a module-load side effect.

**Game.js:**
- Drops `import { setPerpClasses }` and the `setPerpClasses(Game)` call at the IIFE tail.
- Replaces with a side-effect `import './game/perpCtors.js'`.
- Updates two block comments to reflect the new wiring.

**`Database.ts`:**
- Drops `lookupPerpClass` import; adds `perpCtors` and direct `TokenPerp` imports.
- Line 382's `lookupPerpClass(nodeGameType)` → `perpCtors[nodeGameType]`.
- Line 585's known-name `lookupPerpClass('TokenPerp')` → direct `TokenPerp` reference; the now-redundant `if (TokenPerpCtor)` guard collapses to an unconditional `else`.
- The `new TokenPerp(...)` instantiation needs a per-call-site `as ConstructorParameters<typeof TokenPerp>[0]` cast because the legacy config-object shape has `id: string | undefined` fields that violate `exactOptionalPropertyTypes`.

**`DatabasePerp.ts`:** same swap as `Database.ts:382`.

**`GamePerp.ts`:** comment-only updates (header + AniTicker note). Still uses the registry — that's intentional.

**`perpRegistry.ts`:** no API changes. Updated header to document the new wiring; added a comment explaining why `PerpCtor`'s parameter stays `unknown` rather than `GameNodeConfig` (caller-side narrows).

- No new `webxdc.sendUpdate` paths.
- Engine layer untouched (`state.ts`, `materializer.ts`, `LocalEngine.ts`, etc.).
- `addr === state.addr` guard intact.

## Why the registry can't fully go away yet

`GamePerp.ts` is loaded before any of its subclasses (each subclass's `extends GamePerp` requires GamePerp to be defined). If GamePerp.ts directly imports `perpCtors.ts`, the import chain becomes:

```
GamePerp.ts → perpCtors.ts → AgentPerp.ts → GamePerp.ts (mid-load)
```

At step 4, `GamePerp` is undefined and `class AgentPerp extends GamePerp` crashes at module load. The registry's late-binding sidesteps this — `lookupPerpClass(name)` is called at runtime (well after all modules have evaluated), so the registry is fully populated by then.

The seam can fully retire once `GamePerp.BuyPerp`'s dynamic by-name construction moves to a free function in a non-GamePerp module — likely after GameRoot extracts.

## Verification

- `npx tsc --noEmit` — clean.
- `pnpm exec biome check .` — clean.
- `pnpm test` — 624/624 pass.
- `pnpm test:e2e` — 45/45 pass (1 skipped, unrelated).
- `pnpm run build` — both `.xdc` variants build green.

## Test plan

- [ ] CI lint / typecheck / unit-tests / build / playwright all green.
- [ ] Drop the `.xdc` into Delta Chat; smoke-test buy → charge → collect → integrate (Database / Token / Project flows exercise the new lookups).

https://claude.ai/code/session_018qbFtLNgH4p39DmEBtouJu

---

### DIFF
--- a/scripts/game/Database.ts
+++ b/scripts/game/Database.ts
@@ -2,11 +2,10 @@
 // (pending profilesets to integrate) and the integrated TokenPerps.
 // Extracted from scripts/Game.js's IIFE in PR 10 of issue #147.
 //
-// The biggest UI subclass extracted so far (~500 LOC).  GameRoot's surface
-// is narrowed via a local `GameRootForDatabase` interface; the dynamic
-// `Game[node.game_type]` and direct `Game.TokenPerp` lookups are routed
-// through ./perpRegistry.ts so Database can land before every Perp class
-// is extracted in PR 11+.
+// GameRoot's surface is narrowed via a local `GameRootForDatabase`
+// interface.  The dynamic `Game[node.game_type]` lookup is resolved via
+// `perpCtors[name]` (typed direct map, PR 17 of issue #147); the
+// known-name `Game.TokenPerp` reference uses a direct import.
 
 import { getRender } from '../Render.js';
 import appModule from '../app.js';
@@ -23,8 +22,9 @@ import {
 } from './GameNode.js';
 import { OrderedSet } from './OrderedSet.js';
 import { ProfileSet } from './ProfileSet.js';
+import { TokenPerp } from './TokenPerp.js';
 import { mergeData } from './mergeData.js';
-import { lookupPerpClass } from './perpRegistry.js';
+import { perpCtors } from './perpCtors.js';
 
 // ---------------------------------------------------------------------------
 // Local types
@@ -379,7 +379,7 @@ export class Database extends GameNode {
         const node = r.node;
         const nodeGameType = node.game_type;
         if (!nodeGameType) return;
-        const Ctor = lookupPerpClass(nodeGameType);
+        const Ctor = perpCtors[nodeGameType];
         if (!Ctor) return;
         const node_data = groot.getTypeData(bgestalt);
         const perp = new Ctor({
@@ -582,7 +582,6 @@ export class Database extends GameNode {
       });
 
       const triggerQueue: GameNode[] = [];
-      const TokenPerpCtor = lookupPerpClass('TokenPerp');
       // Precompute gestalt-keyed lookup maps so the inner loop is O(M)
       // instead of O(M*N) — a profileset with M tokens against a Database
       // with N children would otherwise scan children.set + all_tokens
@@ -600,13 +599,13 @@ export class Database extends GameNode {
         if (token) {
           // collect update tokens
           update_tokens.push(token);
-        } else if (TokenPerpCtor) {
+        } else {
           // create new tokens
           const type = groot.getType(gestalt);
           if (type && type.game_type === 'TokenPerp') {
             const token_instance = allTokensByGestalt.get(gestalt);
             if (token_instance) {
-              const newToken = new TokenPerpCtor({
+              const newToken = new TokenPerp({
                 id: token_instance.game_id,
                 gestalt: gestalt,
                 path: token_instance.full_path,
@@ -614,7 +613,7 @@ export class Database extends GameNode {
                 renderNodeParent: getFirstId('Database'),
                 ViewMap: getByFirstId('Database'),
                 gameType: type.game_type,
-              }) as TokenPerpLike;
+              } as ConstructorParameters<typeof TokenPerp>[0]) as TokenPerpLike;
               gnode.addChild(newToken);
               triggerQueue.push(newToken);
               new_tokens.push(newToken);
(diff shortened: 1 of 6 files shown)


scripts/game/Database.ts
```diff
@@ -600,21 +599,21 @@ export class Database extends GameNode {
...
+              const newToken = new TokenPerp({
                 id: token_instance.game_id,
                 gestalt: gestalt,
                 path: token_instance.full_path,
                 data: mergeData(type.type_data, token_instance.instance_data || {}),
                 renderNodeParent: getFirstId('Database'),
                 ViewMap: getByFirstId('Database'),
                 gameType: type.game_type,
-              }) as TokenPerpLike;
+              } as ConstructorParameters<typeof TokenPerp>[0]) as TokenPerpLike;
```
experintellia, line 616: Nit (non-blocking): the per-call `as ConstructorParameters<typeof TokenPerp>[0]` cast handles the `exactOptionalPropertyTypes` mismatch between the legacy server-response config shape and `GameNodeConfig`'s strict shape. This is the only direct-import construction site touched in this PR, but as more subclasses retire registry indirection in later PRs they'll all hit the same pattern. Worth keeping an eye on whether a small `makePerp<T>(Ctor, cfg)` helper or a normalized `GameNodeConfigInput` shape becomes worthwhile once two or three more call sites accumulate. No change needed here.

---

experintellia (commented):

## What I verified

**Architecture invariants (re-grepped tree-wide, not just diff):**
- ✅ Engine layer untouched — `state.ts`, `materializer.ts`, `LocalEngine.ts`, `clock.ts`, `webxdc-shim.ts` have zero changes.
- ✅ No `setTimeout` introduced; no new `webxdc.sendUpdate` paths.
- ✅ `delta.addr !== state.addr` guard, `logAction` drop, `resetGame` absence — all still intact.

**#147 quality bar:**
- ✅ Zero `any`, zero `// @ts-*`, zero `!` non-null assertions in any of the changed files.
- ✅ The 10 `as unknown as PerpCtor` casts in `perpCtors.ts` are isolated to one centralized bridge; rationale is documented inline (registry's `unknown`-input parameter shape per `perpRegistry.ts`).
- ✅ One new `as ConstructorParameters<typeof TokenPerp>[0]` cast at `Database.ts:613` — necessary because `exactOptionalPropertyTypes: true` rejects the legacy server-config shape's `string | undefined` fields against `GameNodeConfig`'s strict shape. Well documented in the PR body. Inline note posted on whether to centralize as the pattern recurs.
- ✅ `Game.js` retains its `@ts-nocheck` quarantine header.
- ✅ `lookupPerpClass` survives only at `GamePerp.ts:400` (`BuyPerp`'s by-name construction) — the cycle-bound caller. All non-cycle callers (`Database`, `DatabasePerp`) migrated to direct `perpCtors[name]`.

**Cycle reasoning:** the PR's load-order analysis is correct. `GamePerp.ts → perpCtors.ts → AgentPerp.ts → GamePerp.ts (mid-load)` would crash at the `class AgentPerp extends GamePerp` line. Keeping the registry as a late-binding indirection for that one caller, with `perpCtors.ts` populating it as a module-load side effect (imported by `Game.js` AND consumed by `Database`/`DatabasePerp` for the value), is the right call.

**Module-load order:** `perpCtors.ts` is imported by three call sites (`Game.js` side-effect, `Database.ts` value, `DatabasePerp.ts` value). At least one of those evaluates before any `BuyPerp` flow runs, so the registry is reliably seeded. Fine.

**CI:** lint, typecheck, unit tests (624/624), e2e (45/45), `.xdc` build — all green.

## Stack ordering note

This PR's branch is built atop PR #188 — the diff against `master` would include #188's contents. Please make sure #188 lands first (or merge in stack order). If #188 is rebased during review, #193 will need a rebase too.

## What remains for #147 after this lands

Per the in-issue plan: extract `GameNode.prototype.*` mixins, extract `GameRoot`, type `Render.js` subclasses, then the final cleanup PR (`allowJs: false`, drop the 8 remaining `@ts-nocheck` quarantine markers, close #147). The `perpRegistry` seam itself retires once `GamePerp.BuyPerp`'s by-name construction can move to a free function — likely after `GameRoot` extracts.

---

PR #206, opened 2026-05-06
phase 7 (#147): migrate GameRoot game-values methods (PR 22)

experintellia:
## Summary

PR 22 of issue #147. Migrates 8 game-value methods from `GameRoot.prototype.X = function () {...}` assignments in Game.js into typed methods on the `GameRoot` ES class. **Game.js -184 lines.**

This PR collapses the forward-ref optional setters that PR 21 declared on the class — `setLevel` / `updateStatusBarValues` now resolve to typed direct calls.

## Methods migrated

- **`initGameValues()`** — hydrates value-state from `data.game_values`, anchors first level via `setLevel`, registers GameRoot with the APTicker and starts it. Called once per game-load.
- **`updateGameValues(gv, levelup?, missions?, silent?)`** — partial-delta apply. Each field is fired through its dedicated setter only when present in the payload AND different from the local cache. `ap_snapshot` (authoritative engine AP) is treated as a full reset rather than an additive delta — see preserved comment about issue #153.
- **`useAP(inc)`** — additive AP delta wrapper.
- **`setAP` / `setCash` / `setProfiles` / `setKarma` / `setXP(num?, silent?)`** — write the per-value field, clip to `xp_level.ap_max` / `karma_max` / `profiles_max` / etc., refresh the `data.status_bar.X` slot (val / max / barsize), trigger `renderStatusbar.FXUpdate*(silent)`.

## New types in `GameRoot.ts`

- **`StatusBarSlot`** — single bar's val/max/barsize (+ XP-only level, profiles-only crosssum/tokenslength/tokenslengthmax).
- **`StatusBarData`** — typed shape for `data.status_bar` with the 5 named bars.
- **`RenderStatusbarLike`** — narrow Render Statusbar surface for the `FXUpdate*` fan-out.
- **`GameValuesPayload`** — server-emitted partial-delta shape.
- **`MissionsLike`** — forward-ref for `groot.Missions` (collapses when the Missions singleton API formalises).

## Class fields added

- `profiles_max`, `cash_max`, `ap_offset`, `karma_max` — explicit defaults (`cash_max=10000` keeps the legacy NaN fix; `karma_max=100` matches the legacy hardcoded constant).
- `Missions?: MissionsLike` — set by `loadGame` (still in Game.js).
- `data.status_bar` re-typed `StatusBarData`.
- `renderStatusbar` typed `RenderStatusbarLike`.

## Subtle preserved behavior

- **`setKarma` clipping order** (max-first then negative-max) matches legacy — order matters when `karma_value` could be both > max and < -max (impossible in practice but preserved).
- **`setXP` calls `setLevel` twice** if `xp_value` falls outside the current level both above AND below. Preserves the legacy quirk where the second call only fires if the first didn't fix the range. Both calls are guarded by separate `if`s in legacy and here.
- **`setProfiles` calls `getDBTokensLength()`** (the side-effect variant that mutates `this.DBTokensLength`) before reading the field — order preserved.
- **`updateGameValues` levelup-only side-effects** — `getDatabase().checkNotifications()` + `makeNotifications({ levelup })` only fire when `ap_snapshot` is present AND `levelup === true` AND not silent. All three guards preserved.

- No new `webxdc.sendUpdate` paths.
- Engine layer untouched.
- `addr === state.addr` guard intact.
- Game.js retains its `@ts-nocheck` quarantine header.
- The `// FIXME` and `// TESTING` comments on the legacy methods are preserved verbatim.

## Verification

- `npx tsc --noEmit` — clean.
- `pnpm exec biome check .` — clean.
- `pnpm test` — 624/624 pass.
- `pnpm test:e2e` — 45/45 pass (1 skipped, unrelated).
- `pnpm run build` — both `.xdc` variants build green.

## Test plan

- [ ] CI lint / typecheck / unit-tests / build / playwright all green.
- [ ] Drop the `.xdc` into Delta Chat; smoke-test:
  - [ ] AP regen tick (`APTick` → `setAP`).
  - [ ] Cash payout from a Client collect (`updateGameValues` → `setCash`).
  - [ ] Profile integrate (`setProfiles` updates the crosssum / tokenslength).
  - [ ] Karma incident (`setKarma` clips to ±100).
  - [ ] XP gain → levelup (`setXP` triggers `setLevel` and the levelup notification path).
  - [ ] Existing `tests/e2e/{ap-bar,energy-stat-silent-update,share-merge,sweepstakes-upgrade}.spec.ts` already passing in unit run.

## What remains for GameRoot in #147

Two method-group PRs to follow:
1. **Notifications** — `makeNotifications` (~325 lines), `cueNotification`, `startNotificationQueue`, `uncueNotification`, `compileProvidedKarma`, `openNotification`, `BuyKarma`.
2. **Buy + load + view getters + render hooks** — `BuyPerp`, `refresh`, `loadGame`, `getImperium`, `getDatabase`, `extendRender`, `initEventHandlers`.

Then: retire `perpRegistry`, type Render.js subclasses, final cleanup PR.

https://claude.ai/code/session_018qbFtLNgH4p39DmEBtouJu

---


scripts/game/GameRoot.ts
```diff
@@ -676,11 +739,183 @@ export class GameRoot extends GameNode {
...
+      this.profiles_max = gv.profiles_max;
+    }
+    if (gv.profiles_value !== undefined && gv.profiles_value !== this.profiles_value) {
+      this.setProfiles(gv.profiles_value, silent);
+    }
+    if (gv.cash_value !== undefined && gv.cash_value !== this.cash_value) {
+      this.setCash(gv.cash_value, silent);
+    }
+    if (gv.ap_increment) {
+      this.useAP(gv.ap_increment);
```
experintellia, line 804: Subtle behaviour change worth a sanity check (non-blocking but real):

Legacy: `this.cash_max = gv.cash_max || 10000;`
New: `this.cash_max = gv.cash_max ?? 10000;`

The `||` → `??` swap changes behaviour when `cash_max === 0`: legacy would substitute `10000`, new keeps `0`. The downstream barsize math is `Math.round((cash_value / cash_max) * 120)` — with `cash_max === 0` that's `Infinity * something` → NaN-or-Infinity-clamped through `Math.min(120, Math.max(0, …))`. Probably renders as `120` (full bar) but worth verifying the server bootstrap never sends `cash_max: 0` as a sentinel. If it can, swap to `||` (or `gv.cash_max && gv.cash_max > 0 ? gv.cash_max : 10000` for an explicit guard).

Same pattern is potentially worth checking on the other `gv.X ?? 0` defaults — for those the `?? 0` vs `|| 0` difference is harmless since the fallback equals the falsy value. `cash_max` is the only field where the fallback (`10000`) differs from the falsy value (`0`).

---

scripts/game/GameRoot.ts
```diff
@@ -676,11 +739,183 @@ export class GameRoot extends GameNode {
...
+    // integrateCollected (handlers decrement ap_snapshot but Game.js
+    // never reapplied it pre-#120 follow-up).
+    if (gv.ap_snapshot !== undefined && gv.ap_snapshot !== this.ap_value) {
+      this.setAP(gv.ap_snapshot, silent);
+    }
+    // levelup-only side effects.
+    if (gv.ap_snapshot !== undefined && levelup === true && !silent) {
+      this.getDatabase?.().checkNotifications?.();
+      this.makeNotifications?.({ levelup: this.xp_level.number });
+    }
```
experintellia, line 824: Worth noting: legacy `updateGameValues` calls `this.useAP(gv.ap_increment, silent)` passing `silent`, but the legacy `useAP(inc)` signature accepts only one arg — the `silent` was silently dropped at the legacy layer too. The new code drops `silent` in both the call site here and `useAP`'s signature, which preserves the (always-rendering) effective behaviour. The fact that legacy clearly *intended* AP-increment to honour `silent` (it passes the arg!) is a latent bug that's preserved here. Consider filing a follow-up issue: should `useAP` honour `silent`? In the current state, AP increments via `gv.ap_increment` always re-render the bar even on silent paths, which can be a flicker source.

---

experintellia (commented):

## What I verified

**Architecture invariants (re-grepped tree-wide):**
- ✅ Engine layer untouched — `state.ts`, `materializer.ts`, `LocalEngine.ts`, `clock.ts`, `webxdc-shim.ts` zero changes.
- ✅ No new `setTimeout`. No new `sendUpdate` paths.
- ✅ `delta.addr !== state.addr`, `logAction` drop, `resetGame` absence — all intact.

**#147 quality bar:**
- ✅ Zero `any`, zero `// @ts-*`, zero `!` non-null in `scripts/game/GameRoot.ts`.
- ✅ `Game.js` retains its `@ts-nocheck` quarantine header.
- ✅ Game.js -184 lines as advertised; only the 8 method-block ranges removed.

**Forward-ref optionals collapsed correctly:** the `setAP?(): void` etc. declared in PR 21 are gone; `setLevel` / `APTick` / `updateStatusBarValues` now resolve to direct calls. Clean.

**Migration fidelity (eyeballed against legacy):**
- `initGameValues`: legacy `||` → new `??` — see inline note about `cash_max`.
- `updateGameValues`: every branch preserved 1:1, including the `ap_snapshot` partial-delta semantics, `levelup`-only side effects (3 guards: `ap_snapshot !== undefined && levelup === true && !silent`), and the `// FIXME: TESTING` comment block.
- `useAP`: signature `(inc)` matches legacy; the `silent` arg legacy passed was already a no-op — see inline note.
- `setAP`: `clipped` named the previously-unused `clipap` var; `barsize` now consumes it explicitly. Legacy used `sb.AP.val` (re-evaluated to the same clipped expression) for the calc — equivalent. The "always invoke FXUpdate*" comment for issue #153 is preserved.
- `setCash`: faithful 1:1 migration.
- `setProfiles`: `getDBTokensCrossSum` then `getDBTokensLength()` (side-effect call mutating `this.DBTokensLength`) order preserved, then read of the now-mutated field.
- `setKarma`: clipping order (max-first, then -max) preserved as documented.
- `setXP`: `if (num > xp_value)` greater-only assignment preserved; double `setLevel` for above-max AND below-min preserved (each guarded by separate `if`s); barsize span calc inlined as `span = xp_max - xp_min` for readability — same arithmetic.
- All `renderStatusbar?.FXUpdateX?.(silent)` chains correctly preserve the existing `if (this.renderStatusbar) { this.renderStatusbar.FXUpdateAP(silent); }` semantics.
- Defensive `if (!sb) return;` early-outs replace legacy crash-on-undefined for `data.status_bar` access — wave-norm defensive change.

**Type additions:**
- `StatusBarSlot` / `StatusBarData` tighten the previously-`{ [key: string]: unknown }` `data.status_bar`. Each per-bar slot has the optional fields it actually uses (XP-only `level`, profiles-only `crosssum`/`tokenslength`/`tokenslengthmax`).
- `RenderStatusbarLike extends RenderRootLike` adds the `FXUpdate*` fan-out.
- `GameValuesPayload` makes the partial-delta shape explicit. Every field optional (correct — engine emits partial deltas).
- `MissionsLike` forward-ref keeps `groot.Missions.updateMissions` typed without committing to the eventual full Missions singleton API.
- `APTickerLike` extended with `start(offset?)` and `addListener(node)` for `initGameValues`.

**CI:** lint, typecheck, unit tests (624/624), e2e (45/45 incl. ap-bar / energy-stat-silent-update / share-merge / sweepstakes-upgrade), `.xdc` build — all green.

## Two non-blocking inline notes

- `cash_max ?? 10000` is a real semantic change vs legacy `cash_max || 10000` for the `cash_max === 0` case. Probably never sent by the server, but worth verifying.
- `useAP` legacy receives a `silent` arg from `updateGameValues` that was already silently dropped — preserved here. Latent flicker bug worth filing for a follow-up.

## What remains for #147

Per the PR body: 2 more GameRoot method-group PRs (notifications ~325 lines, then buy/load/getters/render hooks), then `perpRegistry` retirement, Render.js typing, final cleanup PR.

---

PR #214, opened 2026-05-06
phase 7 (#147): extract Render's Set class to scripts/render/RenderSet.ts (PR 29)

experintellia:
## Summary

PR 29 of issue #147 — **first seam in the Render.js wave**.

The Render-side `Set` collection class is essentially the engine-side `OrderedSet` (`scripts/game/OrderedSet.ts`) plus six bulk-action methods used by the renderer's `children` / `decorators` / `cables` / `listeners` / `collisionNodes` collections. Extracted to a new `scripts/render/RenderSet.ts` that **extends `OrderedSet`** with the Render-specific surface (`hide` / `show` / `fade` / `draw` / `removeAll` / `clear`).

## Why this scope

I attempted a wholesale `Render.js → Render.ts` conversion first; it surfaced 1100+ type errors (mostly `TS2683` `this`-implicitly-any from the legacy function-constructor pattern, plus heavy untyped vendor surface from EaselJS / TweenJS / Zynga Scroller / jQuery DOM). That's a multi-day, multi-PR effort.

This PR establishes the **incremental-extraction** approach for the Render wave — same pattern that worked for the GamePerp/GameRoot subclasses. Each subsequent PR pulls one or two classes from Render.js into `scripts/render/*.ts`. Render.js stays `@ts-nocheck`'d until the last extraction lands.

## What changed

**New file: `scripts/render/RenderSet.ts`** (~52 LOC)
- `class RenderSet<T extends RenderNodeLike> extends OrderedSet<T>`.
- Adds the 6 bulk-action methods. Local `RenderNodeLike` forward-ref types the per-node surface (`hide?`, `show?`, `setOpacity?`, `draw?`, `remove?`).
- Optional-chain calls accommodate sets containing items without all methods (e.g. SlowTicker.listeners holds tick objects with only `tick`, not the full Node surface).

**`Render.js`** (still `@ts-nocheck`'d):
- Imports `RenderSet` from the new module.
- Aliases it as `var Set = RenderSet;` so the 5 in-IIFE `new Set()` call sites (DragHandler.collisionNodes, SlowTicker.listeners, Node.children, Node.decorators, Perp.cables) keep working unchanged.
- Drops the inline `var Set = function ...` constructor and the 8 `Set.prototype.X = ...` assignments (~76 lines).

## Architecture invariants — preserved

- No `setTimeout` / `sendUpdate` paths touched.
- Engine layer untouched.
- `@ts-nocheck` on Render.js retained until the wave completes.
[...]

PR #215, opened 2026-05-07
phase 7 (#147): extract Render's DragHandler to scripts/render/RenderDragHandler.ts

experintellia:
## Summary

- Second seam in the Render.js wave (after RenderSet, PR #214). Lifts the window-level `DragHandler` class out of Render.js's IIFE into a fully-typed ES class at `scripts/render/RenderDragHandler.ts` (no `@ts-nocheck`).
- Render.js still carries its quarantine marker; this PR shrinks its inline footprint by ~290 LOC. External call sites (`new DragHandler()`, `node.useDragHandler.…`) keep working through `var DragHandler = RenderDragHandler;`.
- Self-contained: depends on the jQuery `$` global and `OrderedSet` (used directly for `collisionNodes` since the RenderSet bulk-action surface — `hide`/`show`/`fade`/`draw`/`removeAll` — isn't needed here).

## Notes

- Behaviour preserved verbatim except for one piece of dead code: the unused `oldPos` local that the legacy mousemove handler computed but never read.
- `dragVector` is now a per-instance object instead of a prototype-shared one. The original `DragHandler.prototype.dragVector = {}` was a latent quirk (mutations leaked across instances); in practice only one DragHandler exists per Render so the fix is invisible at runtime.
- `DraggableNode` is the minimal contract the handler needs from listening nodes (offsetX/Y, dragStartPos, dragging, getPosition/getSize, trigger, moveTo, clipCablePos). Render's still-untyped `Node` / `Perp` classes already provide all of these.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx biome check` — clean on both touched files
- [x] `npm run test` — 624/624 passing
- [x] `npm run test:e2e` — 45/45 passing (1 pre-existing skip)
- [x] `npm run build` — green

---


experintellia (commented):

Second seam in the Render.js wave. ~290 LOC moved out, 334 LOC of strict TS added.

## What I verified

**Architecture invariants:**
- ✅ Engine layer untouched.
- ✅ No `setTimeout` / `sendUpdate` paths in `RenderDragHandler.ts`.
- ✅ `Render.js` retains its `@ts-nocheck` quarantine — correct for incremental Render wave.

**#147 quality bar:**
- ✅ Zero `any`, zero `// @ts-*`, zero `!` non-null assertions in `scripts/render/RenderDragHandler.ts`.
- ✅ Well-scoped forward-ref interfaces: `DraggableNode` (the minimal contract for listeners), `DragEventLike` (jQuery event shape with `originalEvent.touches` / `changedTouches`), `CollisionPos` return type.
- ✅ Uses `OrderedSet` directly instead of `RenderSet` since `collisionNodes` doesn't need the bulk-action methods (`hide`/`show`/`fade`/`draw`/`removeAll`). Right call — picks the narrower base.

**Migration fidelity (eyeballed against legacy):**
- All legacy `DragHandler.prototype.X` methods migrated 1:1 (init, addListener, removeListener, dragstart, dragend, on, off, trigger, getCollisionPos, testCollisions, getCollisions, mousemove handler).
- Defensive guard added on `removeListener`: `if (listener) listener.trigger('dragend');` — strict-TS norm.
- Constructor throws explicit `Error('RenderDragHandler requires the jQuery global to be loaded.')` if `$` is undefined; legacy crashed with `TypeError`. Improvement.

**Two flagged behaviour changes (both correct):**
- Dead-code drop: legacy `var oldPos = { x: …, y: … };` in the mousemove handler was computed but never read — verified by inspecting legacy `Render.js:378-400` (the next line is `var newPos = { ... node.dragStartPos.x + dm.dragVector.x ... }` which doesn't reference oldPos). Pure dead code.
- `dragVector` per-instance vs prototype-shared: legacy `DragHandler.prototype.dragVector = {}` made all instances share the same mutable object. PR's "only one DragHandler exists per Render" claim is accurate — `new DragHandler()` is called once at IIFE construction. Per-instance is the correct shape; observable behaviour unchanged.

**Render.js side cleanup:**
- `var DragHandler = RenderDragHandler;` alias preserves all in-IIFE `new DragHandler()` and `node.useDragHandler.…` call sites unchanged.
- Publisher entry preserved for any external `getRender().DragHandler` readers.

**CI:** lint, typecheck, unit tests (624/624), e2e (45/45), `.xdc` build — all green.

## Quarantine state

Still 1 marker on `scripts/Render.js`. Render wave continues piece by piece.

## What remains for #147 (Render wave)

Per the PR body's earlier roadmap: foundation classes (Node, SlowTicker), visual primitives (Circle/Text/Sprite/Perp/ButtonInline), decorator family (~7 classes), cables, views (ViewTab/ViewMap/Stage/MainMenu), top-level UI (Popup/Statusbar/StatusItem/DBQueue/MissionPerp/TopscorePerp), then final cleanup PR.

---

PR #217, opened 2026-05-07
phase 7 (#147): extract Render's Text + Sprite primitives to scripts/render/

experintellia:
## Summary

- Fourth seam in the Render.js wave (after `RenderSet` #214, `RenderDragHandler` #215, `RenderNode` #216). Lifts the two leaf visual primitives every remaining Render class either is or extends:
  - **`scripts/render/RenderText.ts`** — `Text` (extends `RenderNode`): div whose innerHTML is the CRLF-normalised text content with a measured-width feedback into the Node bounding box.
  - **`scripts/render/RenderSprite.ts`** — `Sprite` (extends `RenderNode`): div with a sprite-sheet `background-image` indexed by named frames (`frameMap`). Foundation for Perp, PerpSprite, and most FX bling animations.
- Render.js still carries its `@ts-nocheck` marker; this PR shrinks its inline footprint by ~125 LOC.

## How it stays compatible

- `var Text = RenderText;` / `var Sprite = RenderSprite;` aliases keep all in-IIFE call sites working unchanged — `new Text(...)` / `new Sprite(...)` in the FX methods, `extend(Perp, Sprite)`, `extend(PerpSprite, Sprite)`, and the publisher entries.
- Both classes are real ES subclasses of `RenderNode` — `super(config)` runs the registry-allocating `init()`. Subclasses still using `util.extend` (PerpSprite, Perp) reach Sprite's methods through the prototype chain just like with the legacy class.
- Prototype-level defaults (`frameSrc=''`, `frameMap={...}`, `text=''`, `textAlign='center'`) preserved via `static {}` initialiser blocks so the chain-only subclasses pick them up.

## Strictness notes

- `JQueryNodeElem` is now exported from `RenderNode.ts` so subclasses can declare their `jdomelem` matches the parent shape under `exactOptionalPropertyTypes`.
- Zero `any`, zero `// @ts-*`, zero `!` non-null assertions.

## Behaviour-preserved drop of legacy dead code

- The legacy `this._id = _instances.length;` line at the top of both ctors and the `attr('id', 'Sprite' + this._id)` initial id were dead — `RenderNode.init` reassigns `_id` from the registry seam and rewrites the DOM id to `'Node' + _id`. Lines simply absent in the TS versions; observable result identical.
- Text's `ssetSize` typo preserved verbatim (existing call sites rely on the exact name).

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx biome check` — clean
- [x] `npm run test` — 624/624 passing
[...]

PR #216, opened 2026-05-07
phase 7 (#147): extract Render's Node base class to scripts/render/RenderNode.ts

experintellia:
## Summary

- Third seam in the Render.js wave (after `RenderSet` #214, `RenderDragHandler` #215) and the densest one: ~620 LOC of core `Node` methods lifted out of the IIFE into a strict-TS class at `scripts/render/RenderNode.ts`. Render.js still carries its `@ts-nocheck` marker; this PR shrinks its inline footprint by ~615 LOC.
- Foundation for everything that follows in the wave — Sprite, Circle, Text, Perp, Decorator, Cable, Popup, Statusbar, ViewMap all extend `Node` via `util.extend`. With `Node` typed, future subclass extractions can lean on `RenderNode` as a real base.

## Scope

**Moved into RenderNode.ts:** constructor, `init`, `setAttrs`, `onAddInit`, `remove`, `addChild` / `addPopup` / `addDecorator`, `removeChild`, position/offset/center/scaled-position helpers, `moveTo` / `moveBy`, `clipCablePos`, `testParentRadius`, `hide` / `show`, `css` / `updateClass` / `updateRenderProp`, `setZ` / `setOpacity` / `setTransform` / `getTransform`, `setSize` / `getSize` / `getScaledSize`, `getVectorTo` / `getVectorPos`, `draw` / `tick`, `on` / `trigger` / `off`, `setDraggable` / `setClickable`.

**Deliberately left inline in Render.js:** the FX/animation methods (`FXSimple` … `FXElasticTo`). They instantiate `Sprite` / `Text`, which themselves extend `Node` — moving them now would require constructor-injection seams for every visual primitive. They mutate `RenderNode.prototype` via the `Node` alias, so subclasses inherit them through the prototype chain just like with the legacy class. Follow-up after Sprite + Text are extracted.

## Seams

Two minimal injection points keep the new module decoupled from the IIFE state:

- **`setRenderNodeRegistry({ count, add, remove })`** — late-binds to the IIFE's `_instances` / `_ids` / `add` / `remove` so `init` and `remove` can allocate / release `_id`s without importing Render.
- **`setRenderNodeTickers({ tickerRemove, slowTickerRemove })`** — late-binds to CreateJS `Ticker` and the `SlowTicker` singleton so `remove()` can unregister from both.

Both seams are wired in Render.js immediately after the import, before any `new Node(...)` runs.

## Strictness notes

[...]

PR #220, opened 2026-05-07
phase 7 (#147): extract Render's Perp + PerpSprite primitives to scripts/render/

experintellia:
## Summary

Seventh Render seam — the two interactive in-game sprite primitives:

- **`RenderPerp`** (~340 LOC) — the user-draggable perp on the ViewMap. Owns the per-perp drag/click event wiring (frame swaps, decorator hide/show, cable wobble/straighten on drag), random placement (`setRandomPosition`), cable connection helpers (`cableTo`, `cableAnimatedTo`, `cableAnimatedRemove`, `getCableTo`, `getCablesToOrigin`), and the `FXDataIn` / `FXDataOut` data-flow choreography.
- **`RenderPerpSprite`** (~40 LOC) — an inner-sprite child layered onto a Perp's domelem. Reads its frame-map pivots from the parent Perp when its own entries omit them, then snaps to the parent's offset.

Render.js still carries `@ts-nocheck`; this PR shrinks its inline footprint by ~390 LOC. Quarantine count still 1.

## Seams

Cable handling crosses an extraction boundary: `Cable` and `PerpCable` still live in Render.js's IIFE. `RenderPerp.cableTo` constructs a PerpCable, so we late-bind the constructor through `setPerpCableCtor()` — Render.js wires it immediately after `extend(PerpCable, Cable)`. Same pattern as the `setRenderNodeRegistry` / `setRenderNodeTickers` seams from PR #216.

## Constructor-reordering bug caught locally

Legacy Perp's function-constructor called `init` (no `setFrame`), then `initUI`, then upgraded `perpSprite` from config object to `new PerpSprite(…)`, *then* finally `setFrameSrc` / `setFrame` (which calls `draw`). An ES `super()` to `RenderSprite` runs the final `setFrameSrc` / `setFrame` / `draw` block at the end of RenderSprite's constructor — so RenderPerp's `draw` override fires *during* `super`, before the Perp ctor body has had a chance to upgrade `perpSprite`. Since `draw` dereferences `this.perpSprite.updatePosition()`, leaving perpSprite as a plain config object during super crashes with "updatePosition is not a function" and detonates the boot.

**Fix**: pre-upgrade `perpSprite` to a real `RenderPerpSprite` *before* calling `super`, and drop the now-duplicate `setFrameSrc` / `setFrame` calls at the end of the Perp constructor (RenderSprite's super already runs them). PerpSprite's `updatePosition` already guards on `parentNode === undefined`, so the early super-time `draw` is a safe no-op until the Perp ctor body's `addChild(perpSprite)` runs.

Caught by the `collect-icon-after-charge` e2e test on a freshly-restarted Vite dev server — the PR #217 lesson on stale dev-server caching paid off again.

## Other notes

- `*Class` import-alias convention applied to both new modules (`RenderPerpClass`, `RenderPerpSpriteClass`) per the PR #217 IIFE-shadow lesson, even though grep found no `var Perp =` / `var PerpSprite =` collisions — prophylactic.
- `DragHandlerLike` in RenderNode.ts widened to declare `getCollisionPos` and `testCollisions`, which `setRandomPosition` needs. Two-line interface extension.
- `dragBound` overridden as a property initializer (arrow form) rather than a method — matches RenderNode's declared `((pos) => void) | undefined` shape under TS class-shape checks.
- Zero `any`, zero `// @ts-*`, zero `!` non-null in either new file.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx biome check` — clean
- [x] `npm run test` — 624/624 passing
- [x] `npm run test:e2e` — 45/45 passing on a fresh Vite dev server (1 pre-existing skip), incl. the `collect-icon-after-charge` regression that flushed out the constructor-order bug
- [x] `npm run build` — green

---

### DIFF
--- a/scripts/Render.js
+++ b/scripts/Render.js
@@ -9,6 +9,8 @@ import { RenderCircle as RenderCircleClass } from './render/RenderCircle.js';
 import { RenderDragHandler } from './render/RenderDragHandler.js';
 import { RenderNode, setRenderNodeRegistry, setRenderNodeTickers } from './render/RenderNode.js';
 import { applyRenderNodeFX } from './render/RenderNodeFX.js';
+import { RenderPerp as RenderPerpClass, setPerpCableCtor } from './render/RenderPerp.js';
+import { RenderPerpSprite as RenderPerpSpriteClass } from './render/RenderPerpSprite.js';
 import { RenderSet } from './render/RenderSet.js';
 import { RenderSprite as RenderSpriteClass } from './render/RenderSprite.js';
 import { RenderText as RenderTextClass } from './render/RenderText.js';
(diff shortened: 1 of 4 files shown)


experintellia (commented):

Seventh Render seam — the two interactive sprite primitives. ~390 LOC out of Render.js, 531 LOC of strict TS in. The constructor-order bug catch in the PR body is impressive — exactly the class of regression incremental Render extraction is at risk of introducing.

## What I verified

**Architecture invariants:**
- ✅ Engine layer untouched.
- ✅ No new `webxdc.sendUpdate` paths.
- ✅ Four `setTimeout` sites in `RenderPerp.FXDataIn` / `FXDataOut` are cable data-flow choreography (each cable's FXDataOut/FXDataIn delayed by accumulated duration). Preserved verbatim from legacy. Invariant 1 holds.
- ✅ `Render.js` retains `@ts-nocheck`. Quarantine count still 1.

**#147 quality bar:**
- ✅ Zero `any`, zero `// @ts-*`, zero `!` non-null assertions in either new file.
- ✅ Cross-extraction-boundary seam (`PerpCableLike` / `PerpCableCtor` / `setPerpCableCtor`) properly typed and fail-loud (`throw new Error('… seam not wired')`).
- ✅ `RenderNode.ts` change is a 5-line interface widening to add `getCollisionPos` / `testCollisions` to `DragHandlerLike` — minimal surface change.

**Constructor-order fix (the careful bit):**
- The pre-upgrade-perpSprite-before-super pattern correctly addresses the class-field semantics gap: ES `super()` runs `RenderSprite`'s `setFrameSrc/setFrame/draw` before the subclass body executes, and `RenderPerp.draw` dereferences `this.perpSprite.updatePosition()`. Pre-upgrading + the existing `if (!parent) return` guard in `PerpSprite.updatePosition` makes the early super-time draw a safe no-op until `addChild(perpSpriteInstance)` runs after super.
- Comments document the rationale inline. Caught via the `collect-icon-after-charge` e2e test on a fresh Vite dev server — exactly the regression-detection pattern the wave needs.

**Migration fidelity (eyeballed):**
- Legacy short-circuits preserved verbatim with documentation: `clickable: config.clickable || true` and `draggable: config.draggable || true` always evaluate to `true` (legacy quirk). Inline comment notes this.
- `cables: config.cables ?? new RenderSet<PerpCableLike>()` — typed RenderSet now, default constructor preserved.
- `setRandomPosition` preserved including the collision retry loop.
- `cableTo` / `cableAnimatedTo` / `cableAnimatedRemove` / `getCableTo` / `getCablesToOrigin` all migrated 1:1.
- `FXDataIn` / `FXDataOut` chain timings preserved.
- PerpSprite `onAddInit` parent-pivot copy + position-snap preserved; the `if (!frame.pivotx) frame.pivotx = parent.frameMap.normal.pivotx;` defaulting matches legacy 1:1 (including the `Object.prototype.hasOwnProperty.call` check for `pivoty`, distinct from the `!frame.pivotx` falsy-check for `pivotx` — preserved verbatim, asymmetric quirk).

**`*Class` import-alias convention** applied to both new modules per the PR #217 lesson, even though no current `var Perp =` collisions in the IIFE — prophylactic, right call.

**`dragBound` arrow override** — property initializer rather than method matches the declared `((pos: …) => void) | undefined` shape on RenderNode under TS class-shape checks.

**CI:** type-check, lint, unit tests (624/624), `.xdc` build — all green. Playwright e2e still in flight at review time; PR body confirms it passed locally on a fresh Vite dev server (incl. the regression-finding `collect-icon-after-charge` spec). I'll trust the local pass; if CI surfaces a flake the wave's track record makes it likely environmental.

## Quarantine state

Still 1 marker on `scripts/Render.js`. Seven seams down.

## What remains for #147 (Render wave)

Per the working order: decorator family (`Decorator`, `DecoratorReady`, `DecoratorAmount`, `DecoratorLabel`, `DecoratorTimer`, `DecoratorGear`, `DecoratorNew`) → cables (`Cable`, `PerpCable` — these will let the `setPerpCableCtor` seam retire) → views (`ViewTab`, `ViewMap`, `Stage`, `MainMenu`) → top-level UI (`Popup`, `Statusbar`, `StatusItem`, `DBQueue`, `MissionPerp`, `TopscorePerp`) → `SlowTicker` singleton → final cleanup PR.

---

Issue #2, opened 2026-04-28
Audit: dd_app JSON-RPC handler inventory


experintellia:
**Blocker for Phase 3 estimating.** Read the open-source `datadealer/dd_app` Python (Pyramid + MongoDB + Redis + Celery, Artistic 2.0) and produce `docs/handler-map.md`.

## Required output

Markdown table, one row per JSON-RPC handler. Columns:

| Method | File:line | Signature | DB reads (Mongo/Redis/Postgres) | DB writes | Returns | Side effects (Celery, socket emit, etc.) |

## Methods to map (registered in `dd_js/scripts/app.js:140-159`)

`buyPowerup`, `chargePerp`, `collectPerp`, `integrateCollected`, `getPowerups`, `getProvidedPerps`, `getSessionLocale`, `buyKarma`, `buyPerp`, `buySlots`, `getToken`, `loadGame`, `setDisplayName`, `getRanking`, `ping`, `resetGame`, `sellPowerup`, `setPerpCoordinates`, `checkUsername`.

## Also capture

- Any handler that auth-gates on user_id — note which fields will become `webxdc.selfAddr`-keyed.
- Any RNG, anti-cheat, or rate-limiting we'll need to mirror (or deliberately skip).
- Schema of the MongoDB documents each handler reads/writes (will inform the JS state shape in Phase 3 #10).

## Out of scope

Don't port code yet. Pure documentation pass.

Ref planning chat for context.

Issue #1, opened 2026-04-28
Phase 0: project setup, license files, vendor dd_rules data, README


experintellia:
Foundation work, all small, do as one PR.

## Tasks

- [ ] Update `README.md`:
  - "About this fork" section explaining: full-game (not demo) port of Data Dealer to webxdc, fusing `dd_js` (frontend) + `dd_rules` (data) + `dd_app` (backend, ported to in-browser JS).
  - Include this framing quote (planning was done with Claude):
    > webxdc's per-chat sandbox model is almost a better fit for what Data Dealer was trying to be than what they actually built. Their planned multiplayer needed a server, social login, friend graphs — all of which evaporated when the company stopped paying hosting bills. webxdc gives you "every chat group is a game lobby" essentially for free, and there's no service to keep alive. The original architecture is the reason it died; webxdc's architecture is the reason a successor wouldn't.
  - Link upstream repos (datadealer/dd_js, dd_rules, dd_app).
  - Note: planning done with Claude.
- [ ] Add license files at repo root:
  - `LICENSE-CODE.txt` — Artistic License 2.0 (covers dd_js fork + ported dd_app code).
  - `LICENSE-ASSETS.txt` — CC-BY-SA 3.0 Austria (covers dd_rules data, img/, i18n/, fonts).
  - `CREDITS.txt` — original team (Ivan Averintsev, Wolfie Christl, Pascale Osterwalder, Tobi Schäfer, Ralf Traunsteiner) + indication of modifications. Verify existing `LICENSE.txt` doesn't conflict; integrate or replace.
- [ ] Vendor dd_rules data:
  - Copy `default_game.json`, `ruleset_3.de.json`, `ruleset_3.en.json` from `datadealer/dd_rules` into `data/`.
  - Add `data/UPSTREAM.txt` recording the source commit SHA so future updates can be diffed.
- [ ] Tag the pre-conversion `dd_js` HEAD as `pre-pivot/dd_js` so original frontend state is preserved if we need to consult it.

## Acceptance

- README renders with quote, framing, and upstream links.
- Both license files present, `CREDITS.txt` lists all original authors.
- `data/` directory contains three JSON files + UPSTREAM.txt.
- Tag `pre-pivot/dd_js` exists.

Ref: planning chat session.

PR #224, opened 2026-05-07
fix: trigger mission completion when goals auto-complete at seed time

experintellia:
## Summary

A mission whose goals were all flagged complete during seeding — e.g. `buy_perp` goals targeting nodes the player already owned — was left stuck in `active_missions` with every goal checked off but no completion event or rewards. This is the bug visible for "Unofficial Collaboration" (recruit Heidi Plum + Elly DeTelly): both goals 1/1 ✓, mission never finishes.

Two paths needed fixing in `scripts/LocalEngine.ts`:

- `_completeMissionsIfReady` cascade-activated the next mission and ran `_autoCompleteBuyPerpGoals` on its goals, but the completion check had already run earlier in the same call. Now it loops so a freshly activated mission whose goals auto-complete is recognized in the same pass.
- `_repairStuckMissionGoals` only looked at goals where `complete: false`, so it never triggered the completion check when a load-time seed had already flipped every goal to complete. It now also fires when any active mission has all goals already complete, so loadGame recovers stuck saves.

## Test plan

- [x] `pnpm test` — 626/626 pass, including two new regression tests in `tests/handlers/handlers.test.js` using the actual Unofficial Collaboration gestalt
- [ ] Manually load a save that's stuck on this mission and confirm the mission completes and rewards are paid out on next loadGame

https://claude.ai/code/session_01NYADeh7VcvbnkT2bcDZzL8

---
_Generated by [Claude Code](https://claude.ai/code/session_01NYADeh7VcvbnkT2bcDZzL8)_

### DIFF
--- a/scripts/LocalEngine.ts
+++ b/scripts/LocalEngine.ts
@@ -1730,40 +1730,52 @@ function _completeMissionsIfReady(
 ): MissionAdvanceResult {
   var ruleset = _getRuleset();
   var completed: string[] = [];
-  var stillActive = activeMissions.filter(function (mGestalt) {
-    var goals = updatedGoals.filter(function (g) {
-      return g.mission === mGestalt;
-    });
-    if (!goals.length) return true;
-    if (
-      goals.every(function (g) {
-        return g.complete;
-      })
-    ) {
-      completed.push(mGestalt);
-      return false;
-    }
-    return true;
-  });
+  var newActive = activeMissions.slice();
 
-  var newActive = stillActive.slice();
-  if (ruleset && ruleset.missions && completed.length) {
-    _eachMissionDef(ruleset, function (def) {
-      var td = def.type_data;
-      if (!td) return;
-      var req = td.required_mission;
-      var gestalt = td.gestalt;
-      if (!req || !gestalt || newActive.indexOf(gestalt) !== -1) return;
-      if (completed.indexOf(req) === -1) return;
-      var seededGestalt: string = gestalt;
-      newActive.push(seededGestalt);
-      (td.goals || []).forEach(function (g) {
-        updatedGoals = updatedGoals.concat([_seedGoalRow(seededGestalt, g)]);
+  // Loop so a cascade activation whose buy_perp goals auto-complete (because
+  // the player already owns the targets) is itself recognized as finished
+  // in the same pass — otherwise the mission stays in active_missions with
+  // every goal checked off.
+  while (true) {
+    var newlyCompleted: string[] = [];
+    var stillActive = newActive.filter(function (mGestalt) {
+      var goals = updatedGoals.filter(function (g) {
+        return g.mission === mGestalt;
       });
+      if (!goals.length) return true;
+      if (
+        goals.every(function (g) {
+          return g.complete;
+        })
+      ) {
+        newlyCompleted.push(mGestalt);
+        return false;
+      }
+      return true;
     });
-    // Auto-complete buy_perp goals for items the player already owns so a
-    // mission that unlocks after the item was bought doesn't get stuck.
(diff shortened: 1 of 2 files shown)


experintellia (commented):
LGTM (posted as COMMENT — repo blocks self-approval; treat as APPROVE-equivalent).

Reviewing this outside my normal #147 reviewer scope per the user's request. Engine-layer bug fix touching `LocalEngine.ts`'s `_completeMissionsIfReady` cascade and `_repairStuckMissionGoals` startup pass.

## What I verified

**Actual file changes:** only `scripts/LocalEngine.ts` (91 LOC) and `tests/handlers/handlers.test.js` (96 LOC of regression tests). The much larger diff GitHub may show is because the PR branch is based on a very old SHA (`146adbca`, before most of the recent #147 wave); master hasn't drifted on `LocalEngine.ts` or the test file since, so the 3-way merge should be clean. **However, recommend rebasing onto current master before merging** to avoid surprises and keep PR review tooling honest going forward.

**Architecture invariants:**
- ✅ `delta.addr !== state.addr` guard in `state.ts:applyDelta` unchanged.
- ✅ No new `webxdc.sendUpdate` paths.
- ✅ No `setTimeout` introduced.
- ✅ `logAction` drop / `resetGame` absence — both intact.

**Fix #1 — `_completeMissionsIfReady` while-loop:**
- Termination: bounded by total mission count. Each non-break iteration moves at least one mission from `newActive` to `completed`. `newActive.indexOf(gestalt) !== -1` prevents re-activation. Finite. ✓
- Correctness: each iteration finds newly-complete missions, cascade-activates next missions whose `required_mission` matches, runs `_autoCompleteBuyPerpGoals`, then re-iterates so the freshly-activated missions get checked.
- Subtle: legacy ran auto-complete once after all cascades; new runs it once per iteration. Idempotent because the set of owned perps doesn't change during this function (rewards don't seed nodes), so behaviorally equivalent for non-buggy cases and correct for the buggy ones.

**Fix #2 — `_repairStuckMissionGoals` startup re-check:**
- New `hasFinishedActiveMission` predicate: any active mission whose goals (>0) are all `complete: true`.
- Returns `_completeMissionsIfReady(...)` even when `changed === false` if such a mission exists.
- Correctly handles the load-time case where a save was persisted with goals flagged complete but the mission never left `active_missions`.

**Tests (`tests/handlers/handlers.test.js`):**
- Two regressions exercising the actual `HEIDI_MISSION` gestalt:
  1. Removes mission from `active_missions` when all buy_perp goals already complete.
  2. Pays out reward (`cash_value: 100 → 2100`, `xp_value: 0 → 1`) on detected-at-load completion.
- Both go through `loadGame()` — proper integration coverage of the `_repairStuckMissionGoals` invocation path.

**CI:** lint, typecheck, unit tests (626/626 incl. 2 new), e2e (45/45), `.xdc` build — all green.

## Non-blocking observations

1. **Test fixture brittleness** (minor): the `2100` cash and `1` XP assertions are tied to the ruleset's reward configuration for `HEIDI_MISSION` (`16f302f8…`). If that mission's rewards change in the ruleset later, this test breaks in a non-obvious way. Acceptable for a regression test pinned to a specific bug. Worth a comment noting the cash/xp baseline comes from the live ruleset config.

2. **Branch is far behind master** — needs rebase before merging or GitHub's auto-merge UI may give confusing previews. The actual logical changes are clean and won't conflict, but reviewers downstream may waste time on the inflated diff.

3. **One could argue the `_repairStuckMissionGoals` rename is now slightly misleading** — it repairs both stuck goals (the original case, where individual goal `complete` flags need flipping) and stuck missions (the new case, where goals are all complete but the mission row is still active). Consider documenting this in the function header. Tiny doc nit.

## Bottom line

Solid fix, correct logic, appropriate test coverage. Ready to merge after a rebase.

---

PR #229, opened 2026-05-08
fix(perps): swap TokenPerp/SupertokenPerp cableType — closes #191

experintellia:
Closes #191.

## Summary

Legacy `Game.js`'s SuperToken prototype-defaults block at line ~2509 contained a copy-paste typo:

```js
SupertokenPerp.prototype.renderType = 'Perp';
TokenPerp.prototype.cableType = 'inout';   // ← intended SupertokenPerp
```

The misrouted assignment overrode the earlier `TokenPerp.prototype.cableType = 'in'` set in the Token block (line ~2145), leaving runtime in this state:

| Class | Effective `cableType` |
|---|---|
| `TokenPerp`      | `'inout'` (typo override won) |
| `SupertokenPerp` | `'in'` (no override; inherits GamePerp default) |

The intended behavior is the opposite: **`TokenPerp` should be `'in'`** (it's a leaf consumer — cables draw data IN from contained tokens) and **`SupertokenPerp` should be `'inout'`** (it both contains tokens and provides them upward).

## Fix

Two-line value swap:

```diff
 // TokenPerp.ts
-override cableType = 'inout' as const;
+override cableType = 'in' as const;

 // SupertokenPerp.ts
+override cableType = 'inout' as const;
```

Both classes' inline comments now point at this PR + issue #191 instead of preserving the typo.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npm run test` — 626/626 passing
- [x] `npm run test:e2e` — 45/45 passing on a fresh Vite dev server
- [x] `npm run build` — green

⚠️ **Cable visual orientation is hard to assert in e2e without screenshot-diff tooling.** The e2e suite covers buy/charge/integrate flows that exercise both perp types and pass with the swapped values, but doesn't directly assert the cable in/out direction. **Manual visual check in Delta Chat advisable** before release — the issue body warned of "downstream Render code (cable rendering, animation directions) that may have adapted to the buggy values"; if any did, this PR will surface the regression visually.

---
_Generated by [Claude Code](https://claude.ai/code/session_018qbFtLNgH4p39DmEBtouJu)_

PR #228, opened 2026-05-08
fix(game-root): forward silent through useAP — closes #207

experintellia:
Closes #207.

## Summary

Legacy `Game.js` passed `silent` as the second arg to `useAP`, but `useAP`'s signature accepted only one parameter, so the flag was silently dropped at the call boundary. Server deltas with `gv.ap_increment` always re-rendered the AP bar through the FX (Tween-queued) path even on silent paths — a flicker source on partial deltas where `ap_snapshot` and `ap_increment` arrive together (the `ap_snapshot` branch correctly honored `silent`, but the `ap_increment` branch above it did not).

## Fix

Two-line change:

```diff
-  useAP(inc: number): void {
-    this.setAP(this.ap_value + inc);
-  }
+  useAP(inc: number, silent?: boolean): void {
+    this.setAP(this.ap_value + inc, silent);
+  }
```

…plus the matching forward in `updateGameValues`:

```diff
   if (gv.ap_increment) {
-    this.useAP(gv.ap_increment);
+    this.useAP(gv.ap_increment, silent);
   }
```

The latent-bug comment that was sitting on `useAP` since PR #206 is updated to point at this PR as the closing change.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npm run test` — 626/626 passing
- [x] `npm run test:e2e` — 45/45 passing on a fresh Vite dev server
- [x] `npm run build` — green

---
_Generated by [Claude Code](https://claude.ai/code/session_018qbFtLNgH4p39DmEBtouJu)_

PR #230, opened 2026-05-08
doc(game-root): flag active Mission.ts caller in #203 FIXME comment

### LINKED ISSUE #203: latent bug: getOriginGestaltFromOriginTokenGestalt ignores its parameter (hardcoded 'city002' lookup)
State check (2026-05-08): reopening — the bug is **still present** in code despite the issue being closed today.

**Current code at `scripts/game/GameRoot.ts:728-732`:**
```ts
getOriginGestaltFromOriginTokenGestalt(_origintokengestalt: string): string | undefined {
  const origin = Object.values(this.DBOriginTokens).find(
    (ot) => ot.originGameNode?.gestalt === 'city002'
  );
  return origin?.gestalt;
}
```

The implementation is unchanged — parameter still ignored, `'city002'` still hardcoded. A FIXME-style comment was added at lines 714-727 acknowledging the bug and the active caller in `Mission.ts:151`, but no fix was applied.

Acceptance criteria from the issue body remain unmet:
- [ ] Either delete the method or apply fix #2 (`return this.DBOriginTokens[origintokengestalt]?.originGameNode?.gestalt;`).
- [ ] Remove the inline FIXME comment in `GameRoot.ts`.

The fix needs the gameplay-side investigation called out in the inline comment: changing the truthy contract from "city002 has any origin token" to "this profileset IS an origin token" will change which tutorial steps get skipped, so the fix must be paired with a tutorial-step audit.

---

experintellia:
## Summary

Doc-only update to the inline FIXME comment on `GameRoot.getOriginGestaltFromOriginTokenGestalt`. The #203 issue body claims the method has zero callers, suggesting it could be safely dropped. **That claim is stale** — `Mission.ts:151` actively uses this function as a truthiness gate in tutorial step-skip logic:

```ts
if (
  step.integrateProfileSet &&
  groot.getOriginGestaltFromOriginTokenGestalt(step.integrateProfileSet)
) {
  deletefrom = k;
  step.nodelay = true;
}
```

So neither of #203's two fix options is straightforward:

1. **Drop the method** — would break the tutorial step-skip path.
2. **"Fix" the implementation** to honor the parameter — would change the truthy contract from "city002 has any origin token" to "this specific profileset IS an origin token", which is a *different* semantic that would change which tutorial steps get skipped. Whether that's desired or accidental needs gameplay-side investigation.

This PR doesn't fix #203. It just updates the inline FIXME comment so anyone reading `GameRoot.ts` sees the call-site context and the semantic gotcha — useful future context for whoever picks #203 back up after game-logic clarification.

No code path changes; no test impact. Leaving #203 open with this finding.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors (doc-only change)

---

### DIFF
--- a/scripts/game/GameRoot.ts
+++ b/scripts/game/GameRoot.ts
@@ -716,8 +716,19 @@ export class GameRoot extends GameNode {
   }
 
   /** Hardcoded `'city002'` lookup is preserved from legacy.  Suspected
-   *  copy-paste latent bug — should probably read `origintokengestalt`'s
-   *  origin link.  Tracked in issue #203. */
+   *  copy-paste latent bug — name says "given an origin-token
+   *  gestalt, return its origin gestalt", but the implementation
+   *  ignores the parameter and always probes for city002.  Tracked
+   *  in issue #203.
+   *
+   *  ⚠️ Has an active caller: `Mission.ts:151` uses this as a
+   *  truthiness gate in tutorial step-skip logic — a "naive" fix to
+   *  honor the parameter would change the truthy contract from
+   *  "city002 has any origin token" to "this specific profileset IS
+   *  an origin token", which would change which tutorial steps get
+   *  skipped.  The issue body's "zero callers" claim is stale; the
+   *  real fix needs gameplay-side investigation rather than a
+   *  refactor. */
   getOriginGestaltFromOriginTokenGestalt(_origintokengestalt: string): string | undefined {
     const origin = Object.values(this.DBOriginTokens).find(
       (ot) => ot.originGameNode?.gestalt === 'city002'


experintellia (commented):
LGTM (posted as COMMENT — repo blocks self-approval).

Doc-only update on `getOriginGestaltFromOriginTokenGestalt`. Right call to *not* close #203 with a code change here — the investigation finding (active caller at `Mission.ts:151` using the function as a truthiness gate in tutorial step-skip logic) means neither of the issue's two proposed fix paths is straightforward:

1. Drop the method → breaks tutorial step-skip.
2. Honor the parameter → changes truthy contract from "city002 has any origin token" to "this specific profileset IS an origin token", silently changing which tutorial steps get skipped.

The updated FIXME comment captures both findings concisely so anyone reading `GameRoot.ts` (or picking #203 back up) sees the call-site context and the semantic gotcha. ⚠️ marker is appropriate for the "stale claim in issue body" warning.

No code-path changes, no test impact. **CI:** typecheck clean per PR body.

**Branch needs rebase before merge** — same situation as #228 / #229.

---

PR #237, opened 2026-05-08
type-debt: type the addDecorator slot writes via DecoType union

experintellia:
## Summary

`RenderNode.addDecorator` populated a same-name slot on the decorated node keyed by the decorator's `decoType` string, but used a `Record<string, RenderNode>` cast on `this` to do the write. The set of `decoType` values is closed (the six `RenderDecoratorReady` / `Amount` / `Gear` / `Label` / `Timer` / `New` subclasses each pin a literal on their prototype), so the dynamic-key write was load-bearing only because TS lacked the slot field declarations.

- Exports a `DecoType` string-literal union from `scripts/render/RenderNode.ts`.
- Declares the six per-decoType slot fields on `RenderNode` (`DecoratorReady`, `DecoratorAmount`, `DecoratorGear`, `DecoratorLabel`, `DecoratorTimer`, `DecoratorNew`), typed `RenderNode | undefined` to match the `rn.DecoratorGear?.…` read pattern callers already use.
- Narrows `RenderNode.decoType` from `string | undefined` to `DecoType | undefined`, and tightens each subclass's `declare decoType` from `string` to `DecoType` in `scripts/render/RenderDecorators.ts`.
- Replaces the `Record<string, RenderNode>`-cast write in `addDecorator` with a `Pick<RenderNode, DecoType>` alias so the indexed write is type-checked against the declared slot fields.
- Drops the `DecoratorAmount?: DecoratorAmountLike` augmentation in `scripts/render/RenderNodeFX.ts` (it would conflict with the now-declared slot field on the base) and casts through `DecoratorAmountLike` at the one FX call site that needs the `setAmount()` surface.

Read sites elsewhere in the codebase (`rn.DecoratorGear?.setFrame(...)`, `parentNode.renderNode?.DecoratorNew?.remove()`, etc.) keep their existing shape — option (b) was picked over option (a) precisely because that read API is wide.

## Test plan

- [x] `npx tsc --noEmit` — no new errors versus parent branch (only the pre-existing `node:*` / `import.meta.glob` / `@playwright/test` baseline errors)
- [x] `npx biome check` — clean
- [x] `npm test` — 626 / 626 passing (17 test files)
- [x] `npm run test:e2e` — 45 passing, 1 skipped (no failures)
- [x] `npm run build` — green (`dist/scripts/esm-bundle.js 1,098.58 kB`)

https://claude.ai/code/session_01Ah37AyDsgdqiCdLuTeeYDz

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ah37AyDsgdqiCdLuTeeYDz)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ pnpm-debug.log
 /coverage/
 /test-results/
 /playwright-report/
+/.claude/
--- a/scripts/render/RenderDecorators.ts
+++ b/scripts/render/RenderDecorators.ts
@@ -13,7 +13,7 @@
 // the `setRenderDecoratorSlowTicker` injection seam this file
 // used to own — DecoratorTimer imports the singleton directly.
 
-import { type NodeConfig, RenderNode } from './RenderNode.js';
+import { type DecoType, type NodeConfig, RenderNode } from './RenderNode.js';
 import { RenderSlowTicker } from './RenderSlowTicker.js';
 import { RenderSprite, type SpriteConfig, type SpriteFrameMap } from './RenderSprite.js';
 import { RenderText, type TextConfig } from './RenderText.js';
@@ -139,7 +139,7 @@ function getReadyText(mode: 'profile' | 'money' | 'gear'): JQueryDecoratorElem {
 }
 
 export class RenderDecoratorReady extends RenderSprite implements DecoratorBase {
-  declare decoType: string;
+  declare decoType: DecoType;
   offsetToParent: { x: number; y: number };
 
   constructor(config: DecoratorReadyConfig = {}) {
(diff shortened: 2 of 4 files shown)


experintellia (commented):
LGTM. Right shape:

1. New `DecoType` string-literal union exported from `RenderNode.ts` — the closed set of values pinned by the six decorator subclasses.
2. The six per-decoType slot fields declared as class fields (`DecoratorReady?: RenderNode | undefined` etc.) so the dynamic-key write in `addDecorator` is type-checked against declared shape.
3. Each subclass's `declare decoType` narrows from `string` to `DecoType`.
4. `addDecorator`'s `Record<…, …>` cast on `this` replaced with a `Pick`-aliased indexed write — type-checked against the declared slot fields.
5. `RenderNodeFX.ts`'s `DecoratorAmount?: DecoratorAmountLike` augmentation correctly drops (it would conflict with the now-declared base slot) and the one FX call site that needs the `setAmount()` surface gets a localised `DecoratorAmountLike` cast.

Read-site API stays wide (`rn.DecoratorGear?.setFrame(...)` etc.) — option (b) over option (a) is the right call given how many call sites read those slots.

Net cast reduction: the dynamic-key write retires; one FX-side cast added at a single call site. Net negative type debt. CI all green per PR body.

---

Issue #192, opened 2026-05-05
Phase 7: switch Vite from copy-mode to true bundling (minify, tree-shake, sourcemaps)


experintellia:
## Why

After #58 (AMD → ESM end-to-end) and #147 (full strict TS), the codebase finally has the prerequisites for Vite to do real bundling — single module system, ESM imports, type-erased TS — but the build is still in copy-mode rather than producing one minified, tree-shaken artifact. This issue flips that switch.

## Motivation

1. **Tree-shaking only works with ESM bundling.** #58 and #147 set up the conditions; copy-mode leaves the win on the table.
2. **Mobile JS parse time is dominated by file count, not byte count.** #147's strict-TS migration splits each `Game.<Class>` into its own `.ts` file — without bundling the `.xdc` ships dozens of small files that the JS engine parses individually. On mid-range Android this is measurable.
3. **Phase 8 mobile (#80) and the Preact dialog refactor are the deadline.** Both want a fast cold start on phones. Bundling is the cheapest single win.
4. **Bundle size becomes a sane CI metric.** Today, comparing "bundle size before vs. after a change" is impossible — the artifact is many files. After this, a single bundle has a single byte count to track in PRs.

## Current state

- `vite.config.ts` + `vite-plugin-webxdc` produce a `.xdc` containing `scripts/*.{ts,js}`, css, vendor libs, and data files. The .xdc is the deliverable.
- The script files appear to be shipped largely as-is rather than rolled into one or a small number of bundles. The dev server uses Vite's native-ESM dev mode (correct), but prod builds should be doing more.
- No minification, no measurable tree-shaking, no sourcemap upload pipeline.

(Verify exact current behavior before opening the implementation PR — the audit is part of the work.)

## Target state

A production build (`npm run build`) produces:

- **One JS bundle** (or a small handful, e.g. main + a vendor chunk if it materially helps) at `dist/scripts/bundle.[hash].js`. Hash for cache-bust hygiene; not strictly needed inside a `.xdc` but clean.
- **Minified** via esbuild (`build.minify: 'esbuild'`, Vite's default for prod).
- **Tree-shaken** — Rollup's default ESM tree-shaking; verify by spot-checking that an unused exported function from a small module is absent from the bundle.
- **Sourcemaps generated** — `build.sourcemap: true`. Shipped inside the `.xdc` for dev/test builds; **excluded** from release `.xdc` artifacts to save bytes (controlled by build env / mode).
[...]

Issue #80, opened 2026-04-30
Phase 8: make the game mobile friendly


experintellia:
## Overview

The game currently lacks proper mobile support. This issue tracks the work needed to make the game fully playable on mobile devices.

## Goals

- Ensure the layout is responsive and adapts to various screen sizes (phones, tablets)
- Replace or supplement mouse/click interactions with touch events
- Make UI elements (buttons, cards, text) appropriately sized for touch targets
- Prevent unintended zooming or scrolling during gameplay

### Text Readability
- Dialog text must be large enough to read comfortably on small screens (minimum 16px body text, larger for headings)
- Avoid text that gets clipped or overflows dialog boxes on narrow screens
- Ensure sufficient contrast and line-height for legibility

### Buttons & Interactive Elements
- All buttons and clickable actions must have a minimum touch target of 44x44px
- Sufficient spacing between adjacent interactive elements to prevent accidental taps
- Visual feedback (e.g. active/hover states) should work for touch interactions

### Dialogs & Overlays
- Dialogs should be properly sized and centered on mobile viewports
- Content inside dialogs should not overflow or require horizontal scrolling
- Close/confirm buttons must be easily reachable (avoid placing them at screen edges)

## Potential Tasks

- [ ] Add a responsive CSS layout (flexbox/grid, viewport meta tag)
- [ ] Audit and increase font sizes in all dialogs and UI panels
- [ ] Increase button padding/size to meet 44x44px minimum tap target
- [ ] Add spacing between closely grouped interactive elements
- [ ] Replace or augment mouse event listeners with touch event support (`touchstart`, `touchend`, `touchmove`)
- [ ] Test on common mobile screen sizes (360px, 390px, 414px wide)
- [ ] Verify the game loop and interactions work correctly on iOS and Android browsers

## Acceptance Criteria

- Game is fully playable on a modern smartphone in portrait and landscape orientation
- All dialog text is legible without zooming
- All buttons and actions are easily tappable without misclicks
- No horizontal scrolling or overflowing content on small screens

experintellia:
So for mobile, I basically want:
- portrait mode
- new min width is sth. like 300px, like small phone width in portrait mode
- have a break point for smartphone mode.
  - header will have two rows:
  - row 1:
     - logo will be replaced by leveling up bar, leveling up bar gets name on top of it and the star moves to the left.
     - on the right side the orange buttons
   - row 2: the tabs
[...]

PR #244, opened 2026-05-08
Document build pipeline and add bundle size monitoring

experintellia:
## Summary
This PR adds comprehensive documentation of the build pipeline architecture and implements bundle size monitoring in CI. It makes build configuration explicit and visible, with regression guards to prevent accidental bundle bloat or loss of minification.

## Key Changes

- **Architecture documentation** (`docs/architecture.md`): Added detailed "Build pipeline" section documenting:
  - Bundle shape (single IIFE, entry point, output format)
  - Minification strategy (esbuild, explicitly pinned)
  - Tree-shaking verification approach
  - Sourcemap handling (dev/test vs. release builds)
  - Asset inlining thresholds
  - CSS and vendor library loading strategy
  - Static data handling
  - XDC packaging process
  - Background and sequencing context

- **Build configuration hardening** (`vite.config.js`):
  - Added `BUILD_RELEASE` environment variable support to conditionally strip sourcemaps from release builds
  - Explicitly pinned `minify: 'esbuild'` to prevent future Vite major versions from silently disabling minification
  - Explicitly set `sourcemap: !isRelease` to keep sourcemaps in dev/test builds for debugging
  - Explicitly set `assetsInlineLimit: 4096` to make the inlining threshold visible in config
  - Raised `chunkSizeWarningLimit` to 1500 kB to suppress noise from the current ~1.1 MB single-chunk bundle

- **Build regression tests** (`tests/build/build.test.js`):
  - Added minification verification test that checks for long packed lines (average line length > 100 chars)
  - Added sourcemap presence test for dev/test builds

- **CI bundle monitoring** (`.github/workflows/test.yml`):
  - Set `BUILD_RELEASE=1` environment variable on tag builds to strip sourcemaps from release artifacts
  - Added JS bundle size extraction and reporting from the built .xdc file
  - Display bundle size in CI notice and PR comment with both human-readable (IEC) and byte formats
  - Updated PR comment template to include JS bundle size and sourcemap information

## Notable Implementation Details

- Sourcemap stripping is gated by `BUILD_RELEASE=1` environment variable, set only on tag builds by CI
- Bundle size is extracted directly from the built .xdc file using `unzip -p`, avoiding the need to keep unzipped dist/ artifacts
- The minification test uses average line length as a proxy for minification state, since esbuild maintains one statement per line at module boundaries
- All build configuration changes include detailed comments explaining the rationale and linking to relevant issues (#192, #58, #147, #80)

https://claude.ai/code/session_017YJpnkDdgV1cimmqKyJ15S

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -107,6 +107,12 @@ jobs:
         # Casual variant assets are pre-quantized in img-casual/ and
         # icon-casual.png; CI does not need pngquant/oxipng installed.
         # See README.md "Bundle variants" and `pnpm quantize-assets`.
+        #
+        # BUILD_RELEASE=1 on tag builds drops sourcemaps from the .xdc
+        # (vite.config.js: build.sourcemap = !isRelease).  PR / dispatch
+        # builds keep sourcemaps so testers can debug the artifact.
+        env:
+          BUILD_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') && '1' || '' }}
         run: pnpm build:all
 
       - name: Get .xdc bundle info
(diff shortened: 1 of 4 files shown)


experintellia (commented):
LGTM. Out of #147 scope (this is #192) but reviewing per ongoing direction.

## What I verified

**Vite config pinnings** are well-chosen:
- `minify: 'esbuild'` explicit — prevents future Vite-major surprises. The PR comment about "if a future Vite major flips a default" is the right framing.
- `sourcemap: !isRelease` gated by `BUILD_RELEASE=1`. Tag builds drop the .map; PR / dispatch builds keep it for tester debug.
- `assetsInlineLimit: 4096` matches Vite default, pinned for visibility.
- `chunkSizeWarningLimit: 1500` raised from default 500 to suppress noise on the current ~1.1 MB single-chunk bundle. Comment correctly notes that for a webxdc app loaded once from local storage, code-splitting doesn't buy anything (every chunk just becomes another file inside the .xdc) so the single-chunk decision is intentional. Right framing.

**CI workflow change** is correct:
- `BUILD_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') && '1' || '' }}` — only tag builds set the flag. PR builds + workflow_dispatch keep sourcemaps.
- `JS_BYTES=$(unzip -p data-dealer-hq.xdc scripts/esm-bundle.js | wc -c)` extracts size from the built .xdc rather than dist/, avoiding stale unzipped artifacts. Variants share the JS, so HQ-only is sufficient.
- `numfmt --to=iec --suffix=B` for human-readable size; both formats reported.
- `::notice` annotation surfaces the size in the run summary.

**Build regression tests:**
- `avgLineLen > 100` as a minification heuristic — minified esbuild output averages 200-300+ chars/line, unminified Rollup output averages ~30. The 100-char threshold has comfortable margin in both directions. Right call to use avg-line-length rather than total file size or line count, both of which are noisier.
- Sourcemap presence test: only runs `pnpm test` (dev/test branch where `BUILD_RELEASE` is unset). Always expects the .map to ship. Correct.

**Architecture doc** is comprehensive and useful — bundle shape, minification strategy, tree-shaking verification approach, sourcemap handling, asset inlining thresholds, CSS / vendor library loading, static data, .xdc packaging, sequencing context. The tree-shaking-verification recipe ("add an unused export, build, grep the bundle") is good operational documentation that's likely to get used.

CI clean per PR body. No concerns — this is exactly the kind of "make implicit tooling explicit before something silently regresses" hygiene that pays for itself the first time someone bumps Vite.

---

PR #246, opened 2026-05-09
drop jquery-migrate + numeral (~32 KB)

experintellia:
## Summary

Two vendor libs weren't pulling weight — drop them.

### jquery-migrate (13,811 bytes)
Existed to shim removed jQuery 1.x APIs (`$.parseJSON`, `.live()`,
`.bind(event)`, `.size()`, `$.browser`, `$.proxy`, etc.) for code
written against old jQuery. Grepped `scripts/` for every API it
shims — **zero matches**. Pure dead weight.

### numeral + numeral-de (17,924 bytes)
Used at exactly one call site: the `toKSNum` underscore mixin in
`scripts/app.ts`, which called `_.numeral(n).format('0,0')` —
"thousands-grouped integer" with the de-DE locale loaded.
`Intl.NumberFormat('de-DE', { useGrouping: true, maximumFractionDigits: 0 })`
covers it natively and produces the same `1.234` output. Formatter is
cached at module scope (one instance, not one per call).

### Changes
- `index.html` — removed three `<script>` tags
- `scripts/app.ts` — replaced `toKSNum` body with `Intl.NumberFormat`,
  dropped the `numeral: numeral` mixin entry, updated the vendor-libs
  comment
- `types/env.d.ts` — dropped `NumeralStatic` and the `var numeral` /
  `_.numeral(...)` declarations
- `scripts/vendor-install.ts` — removed `jquery-migrate` and
  `numeral@1.4.5` from the npm install list, the corresponding `cp(...)`
  lines, and the version-notes comment for jquery-migrate
- `tests/toolchain.test.js` — moved the three filenames from `required`
  → `removed`
- Deleted `vendor/jquery-migrate.js`, `vendor/numeral.js`,
  `vendor/numeral-de.js`

### Bytes saved
| file | bytes |
|---|---|
| `vendor/jquery-migrate.js` | 13,811 |
| `vendor/numeral.js` | 17,107 |
| `vendor/numeral-de.js` | 817 |
| **total raw** | **31,735** |

Less compressed inside the .xdc zip.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — clean
- [x] `pnpm test` — all 639 tests pass (including the toolchain
      vendor-file assertions, now updated)
- [x] `pnpm build` — succeeds; smoke-checked `data-dealer-hq.xdc`
      contains no `jquery-migrate` or `numeral*` entries under `vendor/`

---
[...]

PR #247, opened 2026-05-09
strip underscore: extract mixins + native swaps

experintellia:
## Summary

Drops `vendor/underscore.js` (~42 KB on disk, ~8 KB inside the zipped
.xdc) and the `<script>` tag that loaded it.

An audit (`grep -rEoh '_\.\w+' scripts/ --include='*.ts'`) showed the
codebase's `_.foo(...)` calls were dominated by the project-local
helpers registered in scripts/app.ts (`toKSNum`, `sprintf`, `span`,
`pad0`, `crlf2html`, `toTime`, `renderView`, `game`, `_`, `__`,
`numeral`); the only real underscore APIs in use were `template`,
`shuffle`, `debounce`, plus the few `each` / `range` / `contains` /
`filter` calls embedded inside view templates.

## What changed

A new `scripts/dd-helpers.ts` (~250 lines):

- Exports each former mixin as a real ESM function — callers now
  `import { toKSNum, sprintf, span, ... }` directly instead of reaching
  through `globalThis._`.
- Provides a `templateHelpers` namespace with `each` / `range` /
  `contains` / `filter` shims so the view templates' inline
  `_.foo(...)` calls keep working unchanged. **No `views/*.html`
  was edited** — every template under views/ continues to use the same
  `<% %>` / `<%= %>` source.
- Replaces `_.template(text, null, { variable: 'D' })` with a
  ~30-line `compileTemplate(text)`. Verified the templates only use
  `<% %>` and `<%= %>` (`grep -rn '<%-' views/` returns nothing), so
  the impl is intentionally minimal.
- Inlines `_.shuffle` / `_.debounce` natively; the only other real
  underscore usage (`_.clone(tokens)`) was already a manual `slice()`
  in scripts/game/ProfileSet.ts.
- One-call registration hook (`registerTemplateHelpers`) is invoked
  from app.ts (`renderView`, `game`, `gettext`, `ngettext`) and
  Render.ts (`RenderSprite`, `RenderAmount`) — replacing the two
  `_.mixin({...})` blocks.

## Decision: option 2 on `_.template`

Per the task description, two paths existed for `_.template`:
1. Keep the 42 KB vendor lib loaded just for the template compiler.
2. Replace it with a tiny in-tree implementation.

Picked option 2. Templates use only `<% %>` and `<%= %>` (no `<%-`,
no settings other than `variable: 'D'`), so a 30-line
`new Function('D', '_', '__esc', body)` impl covers everything
already in views/. The 42 KB vendor lib for one feature was not worth
keeping.

## Files touched

`scripts/`:
- `dd-helpers.ts` (new) — helpers + template engine
- `app.ts` — uses `compileTemplate`; registers `renderView` / `game`
  / `_` / `__`; drops the `_.mixin({...})` block
- `Render.ts` — registers `RenderSprite` / `RenderAmount` via
  `registerTemplateHelpers`
- `Game.ts` — `_underscore().shuffle` → `shuffle` (helpers)
- `i18n.ts` — `_.sprintf` / `_.toKSNum` → direct imports
- `vendor-install.ts` — drops underscore from npm install + cp list
- `game/ClientPerp.ts`, `game/ContactPerp.ts`, `game/Database.ts`,
  `game/GameRoot.ts`, `game/ProjectPerp.ts`, `game/TokenPerp.ts` —
  `globalThis._.{toKSNum,sprintf,span,debounce}` → named imports
- `render/RenderText.ts`, `render/RenderDecorators.ts`,
  `render/RenderTopLevelUI.ts`, `render/RenderNodeFX.ts`,
  `render/RenderPerp.ts` — same; also drops the local
  `getUnderscore()` helpers and `UnderscoreXxxLike` interfaces

`tests/`:
- `build/build.test.js` — asserts `vendor/underscore.js` is *not*
  shipped (moved from `required` to `removed`); index.html no longer
  references it
- `toolchain.test.js` — same vendor-files assertion split

`types/env.d.ts`:
- Drops the `UnderscoreStatic` interface and `var _: UnderscoreStatic`
  global declaration

`vendor/`:
- `underscore.js` deleted

`index.html`, `vite.config.js`:
- `<script src="vendor/underscore.js">` removed; vendor-libs comment
  updated

## Bundle byte counts

|                       | Before        | After         | Δ        |
|-----------------------|---------------|---------------|----------|
| `vendor/underscore.js`| 42,472        | (deleted)     | −42,472  |
| `scripts/esm-bundle.js` | 1,100,914   | 1,101,828     | +914     |
| `data-dealer-hq.xdc`  | 15,145,172    | 15,137,116    | −8,056   |

The .xdc savings (~8 KB) is smaller than the on-disk delta (~42 KB)
because vendor/underscore.js compresses well inside the .xdc's zip.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — clean
- [x] `pnpm test` — 639 / 639 pass
- [x] `pnpm build` — succeeds; .xdc shrinks per the table above
- [ ] `pnpm test:e2e` — Playwright suite touches templates and is the
      real regression net here; **not run in this sandbox** (no
      browser available). Run before merge.
- [ ] Smoke-test: load `data-dealer-hq.xdc` in Delta Chat Desktop and
      walk the golden-path flow, especially anything that hits a
      template render (popups, profile-set queue, statusbar). Not run
      in this sandbox.

---
_Generated by [Claude Code](https://claude.ai/code/session_017a287EC97uvMeGtYehSvWz)_


experintellia (commented):
CI green, diff matches the prompt. One real bug, one merge-order note.

## Bug: `toKSNum` will leak fractional digits

`scripts/dd-helpers.ts`:
```ts
const KSNumFormat = new Intl.NumberFormat('de-DE');
```

`Intl.NumberFormat('de-DE')` defaults to `maximumFractionDigits: 3`, so any non-integer input would now render with up to three decimals. The old `_.numeral(n).format('0,0')` rounded to integer. PR #246 has this right — `{ maximumFractionDigits: 0 }`. Same fix here:

```ts
const KSNumFormat = new Intl.NumberFormat('de-DE', { maximumFractionDigits: 0 });
```

Current call sites probably all pass integers, but the contract changed silently and the next contributor wiring `toKSNum` to a fractional input gets unexpected output.

## Merge-order with #246

This PR and #246 conflict on `app.ts`, `index.html`, `vendor-install.ts`, `env.d.ts`, `tests/toolchain.test.js`. Suggest landing **#246 first** (smaller blast radius), then rebasing this — at which point the `numeral` passthrough function (`dd-helpers.ts:58-64`) and the `templateHelpers.numeral` entry should also be dropped, since `globalThis.numeral` will be gone. Your own comment notes "none of the migrated templates use it directly any more," so removal is safe.

## Nice work

- `templateHelpers` namespace + `_missingHelper` warner is a tidy way to keep the 60+ view files unchanged while late-binding the four entries that need a closure (`renderView`, `game`, `RenderSprite`, `RenderAmount`).
- ~30-line `compileTemplate` is right-sized for the `<%= %>` / `<% %>` surface actually exercised; the `<%- %>` escape path is wired but unused, fine.
- `i18n.ts` migration to direct `import { sprintf, toKSNum }` is the right cleanup.

Non-blocking nits (your call):
- `getJQuery()` at the bottom is scope creep (PR is about underscore, not jQuery). Six lines, harmless — leave or split.
- `RenderSprite: (...args: any[]) => string` uses `any[]` at the namespace level. Pragmatic given the concrete impls narrow at registration time. Comment already explains it.

---

experintellia:
Pushed `ad73b8a`:

- **Bug fix**: `KSNumFormat` now passes `{ maximumFractionDigits: 0 }` to `Intl.NumberFormat`, matching the legacy `_.numeral(n).format('0,0')` integer-only contract. Tests still 639/639.
- **Nit**: dropped `getJQuery()` (unused, scope creep).

Left alone:

- **Merge-order with #246**: agree #246 should land first; will rebase here and drop the `numeral` passthrough + `templateHelpers.numeral` entry once `globalThis.numeral` is gone.
- **`RenderSprite: (...args: any[])`** kept as-is per your nod — comment already explains the contravariance trade-off.

---

experintellia (commented):
Both items addressed cleanly:

- `KSNumFormat = new Intl.NumberFormat('de-DE', { maximumFractionDigits: 0 })` — matches #246, and the new comment explaining the `'0,0'` legacy contract is the right level of detail.
- `getJQuery` (and its `JQueryStatic` type-only import) removed.

CI green. LGTM on this round.

Reminder for after #246 lands: rebase, then drop the `numeral` passthrough function in `dd-helpers.ts` and the `templateHelpers.numeral` entry — `globalThis.numeral` will be gone.

---

PR #270, opened 2026-05-12
build: harden test/typecheck gating + enable .ts test discovery

experintellia:

## Summary

Four small, surgical fixes to the build/test infrastructure that were silently letting bugs through:

| # | Sev | Fix |
| --- | --- | --- |
| 1 | HIGH | `vitest.config.js` include glob extended to `.ts`/`.tsx`/`.mts`; `tests/e2e/**` excluded. Any `.test.ts` file was previously silently skipped. |
| 2 | HIGH | `lefthook.yml` gains a `pre-push` stage running `pnpm typecheck` and `pnpm test`. Pre-commit stays Biome-only. |
| 3 | HIGH | `playwright.config.ts` reads `BUILD_VARIANT` from env (default `hq`, validated against `hq|casual`) and forwards it to the dev-server env. CI can now matrix both variants. |
| 4 | MED | New `tsconfig.tools.json` (+ `pnpm typecheck` runs both) brings `vite.config.js`, `vitest.config.js`, `playwright.config.ts`, and `tools/quantize-assets.mjs` under typecheck. |

## Test plan

- [x] New `tests/toolchain.test.ts` proves the vitest glob change works (`1 + 1 === 2`; would be invisible without the config fix).
- [x] `pnpm typecheck` runs both tsconfigs; passes.
- [x] `pnpm test` — 625 tests, 18 files (was 17 + 624).
- [x] `pnpm exec lefthook run pre-push` — typecheck + test both succeed locally.
- [x] `BUILD_VARIANT=casual pnpm test:e2e` boots the casual build (verified locally; CI workflow update left as follow-up).

## Notes

- One JSDoc cast in `vite.config.js` (Rolldown union narrowing) was required so `checkJs` accepts existing `.output` access; alternative was excluding `vite.config.js` from tools-tsconfig coverage, which defeats the goal.
- CI workflow not modified — operator can add a `BUILD_VARIANT` matrix to the Playwright job to actually exercise casual on every push.

---


tests/toolchain.test.ts
experintellia, line 1: how exactly does this make us notice it?

experintellia: Good catch — the bare `.test.ts` canary was silent in exactly the regression case it was supposed to guard. If vitest stops picking up `.test.ts`, this file just doesn't run and the test count quietly drops by one; nothing fails.

Closed the gap from the always-discovered `.js` side in 38f6b8f. `tests/toolchain.test.js` now:

1. Reads `vitest.config.js` and asserts the include glob still names `ts`/`tsx`/`mts`.
2. Asserts the companion `tests/toolchain.test.ts` file exists on disk.

Either pre-condition failing surfaces as a loud test failure. Verified by temporarily dropping `ts` from the glob locally — got `FAIL  tests/toolchain.test.js > vitest TypeScript test discovery > vitest.config.js include glob covers .ts/.tsx/.mts`. Restored, passes.

---

experintellia (commented):
## Review — build/test gating + .ts test discovery

**Overview.** Four infra-only fixes: vitest discovers `.ts` tests, lefthook gains a `pre-push` typecheck+test gate, Playwright validates and forwards `BUILD_VARIANT`, and a `tsconfig.tools.json` brings config files under `checkJs`. Surgical and contained (+71/−4 across 7 files).

### Strengths
- `vitest.config.js` exclude correctly carves out `tests/e2e/**` — confirmed all e2e files use `*.spec.ts` which would otherwise be slurped in by the new glob.
- `BUILD_VARIANT` validator in `playwright.config.ts` matches the existing guard in `vite.config.js:21-23`, so the two configs agree on the allowed set.
- The `checkJs` workaround in `vite.config.js` (double cast through `any`) is ugly but pragmatic and well-commented; pulling in `rolldown` types as a real dep would be worse.
- Smoke test `tests/toolchain.test.ts` is intentionally trivial — it exists to fail if the vitest glob regresses, which is the right scope.

### Issues / suggestions
- `tsconfig.tools.json` sets `"types": ["node"]` but `@types/node` is not asserted in the PR. If it isn't already a devDependency the second `tsc` invocation will fail in fresh installs — worth a one-line check before merge (`grep '@types/node' package.json`).
- `pre-push` runs the full `pnpm test` (625 tests). Acceptable but worth flagging: developers who push frequently will feel it. A `--changed` or `--bail` flag, or splitting into a fast vs. slow stage, could be a follow-up.
- The toolchain smoke test asserts `1 + 1 === 2`. Consider also `import.meta.url.endsWith('.ts')` or a `typeof` check that only compiles under TS — a JS-renamed file would still pass `1+1===2`. Minor.
- `BUILD_VARIANT || 'hq'` treats empty string as default; that's fine, but `??` would be more precise. Tiny nit.
- PR notes "CI workflow not modified" — the variant-matrix promise is unfulfilled until someone adds the matrix to `.github/workflows/*`. Worth tracking as an issue, not a blocker.

### Risk
Low. Pure infra; no runtime behavior change. The only way this breaks prod is if `pnpm typecheck` now fails on `tsconfig.tools.json`, which the author claims is green locally.

**Verdict:** LGTM once `@types/node` presence is confirmed.

---

experintellia:
Thanks for the careful review. Addressed in 52c856b:

**1. `@types/node` confirmed** — already a devDependency at `package.json:24` (`"@types/node": "^22.0.0"`), so the fresh-install path works.

**2. TS-only canary** — `tests/toolchain.test.ts` now anchors on `import.meta.url.endsWith('.test.ts')` AND uses an `as const` TS-only assertion. A `.js` rename or a regression that disables the TS loader for tests fails to parse rather than silently passing `1+1===2`.

**3. `BUILD_VARIANT || 'hq'` → `?? 'hq'`** — matches the audit-wide `||` → `??` sweep (same nit flagged by the render PR review).

**Acknowledged but not addressed:**

- **`pre-push` runs full `pnpm test`.** Agreed — splitting into a fast/slow stage or `--changed` is a fair follow-up but out of scope for this PR.
- **CI matrix for `BUILD_VARIANT`.** PR ships the knob; the workflow change is intentionally a follow-up as the PR body notes.

---

tests/toolchain.test.ts
experintellia: well do we need it in the pr then? if it was just used for loacal testing?

experintellia: Fair pushback — let me defend the file's value because I think it's not redundant with the .js side, though I should have made that clearer.

The two halves catch different regressions:

- **`tests/toolchain.test.js`** verifies *static configuration*: vitest.config.js still names `.ts` in its include glob, and a .ts canary file exists on disk. It is itself a `.js` file (so always discovered), but it cannot observe whether the TypeScript pipeline actually *works*.

- **`tests/toolchain.test.ts`** is the live proof that the pipeline *does* work end-to-end. If the glob is correct but esbuild's TS transform breaks (e.g. a vitest upgrade ships a regression, or someone adds a `transform: false` for `.ts` files), this file fails to parse/load and vitest reports it as a failed test file. The .js side cannot catch that — it only inspects config, not behaviour.

I tightened the .ts file in 52c856b so it ONLY passes via TS-specific behaviour:

```ts
expect(import.meta.url).toMatch(/\.test\.ts(\?|$)/);
const literal = [1, 2] as const;
```

The `as const` is TS-only syntax that esbuild must strip; a JS loader rejects it. The `import.meta.url` assertion catches a `.test.js` rename. So a passing run of *this specific test* genuinely proves "vitest can find a .test.ts file AND the TS loader is wired up AND `as const` strips correctly." That's a real signal, not local-only.

If you'd rather drop it and rely on the static checks alone, I can — but I'd lose coverage of the TS-pipeline regression class.

---

PR #271, opened 2026-05-12
fix(core): boot() unconditional, _persistDelta dual-path, showFatal XSS, shim observability

experintellia:

## Summary

Five fixes in the core/bootstrap layer, scoped to behavior that's actually observable.

| # | Sev | File | Fix |
| --- | --- | --- | --- |
| 1 | CRITICAL | `scripts/bootstrap.ts` | Always call `boot()`. The `typeof webxdc !== 'undefined'` guard left `_currentState` null whenever the shim was absent (Node, tests without setup), so the first `getState()` in `continueStart` threw. `boot()` already handles the no-webxdc case (seeds `freshState('')`, skips `setUpdateListener`). |
| 2 | HIGH | `scripts/LocalEngine.ts` | `_persistDelta`: when the test `_sendDelta` spy is set, skip `webxdc.sendUpdate`. Previously both fired and the shim re-applied via its listener → double-mutation. JSDoc on `setSendDelta` documents the new short-circuit so future round-trip tests don't install the spy by accident. |
| 3 | HIGH | `scripts/bootstrap.ts` | `showFatal` used `$.html(...err.message...)` — XSS via error message. Now sets the headline via `.html()` for the literal `<br>`, then assigns the detail via `<small>`'s `textContent`. |
| 4 | MED | `scripts/i18n.ts` | `String.prototype.substr` → `slice`. Deprecated API; behavior preserved. |
| 5 | MED | `scripts/webxdc-shim.ts` | `localStorage` read/parse/write `catch (_) {}` now `console.warn` with the error. Quota/parse/private-mode failures are observable. Parse failure also calls `localStorage.removeItem` so the next reload doesn't replay the same warning until a successful `_persist` overwrites it. |

## Tests added

- `tests/bootstrap.test.js` — boot path without `globalThis.webxdc`: asserts the source-level guard is gone and that `boot()` with no webxdc seeds a valid state synchronously. Uses `vi.resetModules()` so the assertion is order-independent.
- `tests/handlers/persist-delta.test.js` — with `_sendDelta` set, `webxdc.sendUpdate` is not called.

## Dropped from the original scope

- **PRNG seed derivation from selfAddr.** The original audit overstated the severity. `_generateId` prefixes IDs with `Date.now()`, so the constant `0xdeadbeef` seed only mattered for two peers generating IDs in the same millisecond. Each peer owns its own `state.nodes_charging` slot and peer-derived state is namespaced by `selfAddr` (`state.peers[addr]`), so cross-peer collect_id "collisions" never share an ID namespace. Reverted to the historical constant. (There is one related edge case — same-user multi-device simultaneous taps — being tracked as a follow-up issue.)
- **`resetPrngSeed` export.** Was an API surface added on top of the seed-derivation change. With the derivation reverted, this had no callers.

## Test plan

- [x] `pnpm typecheck` clean.
- [x] `pnpm test` — 661 passing.
- [x] Each commit verified individually with both gates.
- [x] `_persistDelta` keeps a synchronous `setState(applyDelta(...))` after the spy fires to preserve the listener-echo invariant the `listener-echo.test.js` suite relies on.

## Notes

- All changes verified safe for webxdc state replay — the PRNG is only ever invoked from inside handler bodies (`collectPerp`), never from `applyDelta` / `state.ts` / `boot.ts` / `materializer.ts`. The replay path consumes precomputed IDs from delta payloads; it does not re-roll.


experintellia (commented):
## Review — core/bootstrap audit fixes

**Overview.** Six fixes to the bootstrap/core layer: unconditional `boot()`, `_persistDelta` no longer double-echoes when a test spy is installed, PRNG is per-peer-derived instead of the constant `0xdeadbeef`, XSS in `showFatal`, `substr → slice`, and observable warnings in the webxdc-shim storage paths. +245/−15 across 8 files.

### Strengths
- The bootstrap fix is clean: removing the `typeof webxdc !== 'undefined'` guard and trusting `boot()` to handle the no-webxdc case is the right shape. The source-text assertion in `tests/bootstrap.test.js` is brittle on purpose — appropriate.
- `_deriveBootSeed` mixes `_djb2(addr) ^ Date.now() ^ (addrHash << 16)` — XOR-shift across the word spreads entropy reasonably. Re-using existing `_djb2` avoids new code paths.
- Lazy seeding via `_ensurePrngSeeded()` neatly avoids the circular `boot.ts → LocalEngine.ts` wiring the PR description calls out.
- XSS fix in `showFatal` is correct: hardcoded HTML for the wrapper, `textContent` for the user-influenced message. No element-removal race because the `.html()` write is synchronous.
- `_persistDelta` early-return preserves the listener-echo invariant by still applying `setState(applyDelta(getState(), delta))` — good attention to detail; without it the spy-installed tests would see un-mutated state.

### Issues / suggestions
- `_deriveBootSeed('')` reduces to `Date.now()` only. The lazy path may run before `webxdc.selfAddr` is populated, in which case two peers booting in the same ms produce identical streams — exactly the bug this PR fixes. Worth either (a) re-seeding once `selfAddr` becomes known, or (b) folding `performance.now()` / `crypto.getRandomValues` into the empty-addr fallback. The PR description claims "production boot wires `resetPrngSeed()` once selfAddr is known" but I don't see that wiring in the diff.
- `tests/bootstrap.test.js:42`: `if (!getBootPromise()) boot({ selfAddr: '' });` relies on shared module-level cache from prior tests in the suite. That makes the test order-dependent — it'll pass when run with the full suite, fail in isolation (`vitest run tests/bootstrap.test.js`). The comment acknowledges this but a `vi.resetModules()` setup would be cleaner.
- `_ensurePrngSeeded` is called on every `_rng()`. Trivially cheap but the `var s = _prngSeed as number` then writing back `_prngSeed = s` round-trips the same value through a local for type narrowing — fine, just noting.
- `resetPrngSeed` is now part of the exported API surface (in `app.ts` and the `LocalEngine` aggregate). It's documented as test-only but production callers can reach it. Consider a `__testing__.resetPrngSeed` namespace, or at least an `@internal` JSDoc tag, to make the intent obvious to consumers.
- `_persistDelta` change: when both `_sendDelta` spy AND a real webxdc shim are installed, the shim's listener will no longer fire. Any future test that wants to assert "the listener actually round-trips" must remember not to set `_sendDelta`. Worth a comment in `setSendDelta` itself, not just `_persistDelta`.
- Minor: `webxdc-shim.ts` parse-error path resets `_updates = []` but never re-persists, so the malformed key lingers in localStorage. The next `_persist()` will overwrite it on the first new update, but a `localStorage.removeItem(STORAGE_KEY)` on parse failure would be cleaner.

### Risk
Medium. The PRNG and `_persistDelta` changes touch hot paths. The PRNG seed change is observable to any test that depends on specific generated IDs — the PR claims test determinism is preserved via explicit `setPrngSeed`, but it's worth a careful grep for any test that runs without calling `setPrngSeed` and depends on a specific id value.

**Verdict:** LGTM with two follow-ups: (1) confirm whether production explicitly calls `resetPrngSeed(selfAddr)` once it's known (the PR body says yes, the diff doesn't show it), and (2) consider isolating the bootstrap test from shared module state.

---

experintellia:
Thanks for the thorough review. Addressed in 6e568e1:

**1. PRNG empty-addr entropy (the substantive one).** You were right — the production wiring you couldn't find isn't there: only the lazy `_ensurePrngSeeded` runs on first `_rng()`, which reads `webxdc.selfAddr` at that moment. In practice `_generateId` is only called from `collectPerp` (handler-time, post-boot, so selfAddr is populated), but the empty-addr fallback was indeed `Date.now()`-only. `_deriveBootSeed('')` now mixes in a word from `crypto.getRandomValues` when available, with a graceful fall-back to pure `Date.now()` on older runtimes. Two peers booting in the same ms with no selfAddr now pick disjoint streams.

**2. Bootstrap test order-independence.** `tests/bootstrap.test.js` now calls `vi.resetModules()` in `beforeEach` and re-imports `boot.js` per test, so it passes both in-suite and in isolation (`vitest run tests/bootstrap.test.js`). Verified both invocations locally.

**3. `@internal` on `resetPrngSeed`.** Added JSDoc tag plus a NOT-intended-for-game-logic note. The remote-namespace approach (`__testing__.resetPrngSeed`) would require restructuring the LocalEngine aggregate export, which feels out of scope for this PR — open to doing it in a follow-up if the reviewer feels strongly.

**4. `setSendDelta` comment.** Added JSDoc to `setSendDelta` itself (not just `_persistDelta`) calling out that installing the spy short-circuits `webxdc.sendUpdate`. Future round-trip tests now have a discoverable warning.

**5. webxdc-shim parse-error path.** Now `localStorage.removeItem(STORAGE_KEY)` after the malformed history is dropped, so the next reload doesn't replay the same warning until a successful `_persist` overwrites it.

**Acknowledged but not changed:**

- **PRNG test grep.** I grep'd for `setPrngSeed` and `resetPrngSeed` callers in test files — the only deterministic-id-dependent tests already call `setPrngSeed(N)` explicitly. Tests that don't assert on specific generated values are unaffected.
- **`_rng()` round-trip via local var.** Agreed it's a noise pattern; the `as number` narrow could land in a typing follow-up rather than touch the hot path.

---

Issue #10, opened 2026-04-28
Phase 3: state model + webxdc-update persistence


experintellia:
**Blocker for all other Phase 3 issues.** Defines the JS state shape and how multi-device sync works.

**⚠️ Blocked on #31 (toolchain swap). New modules in this PR are written as ESM, which only works once #31 lands. Don't start until #31 merges.**

## Storage architecture (MVP)

- **Single source of truth: `webxdc.sendUpdate` history.** Every state mutation produces a delta update. On boot, `setUpdateListener(cb, 0)` replays the full history; the engine builds current state in memory.
- **No local cache for MVP.** Replay-from-zero on every cold start. Acceptable until history grows large; optimization tracked separately as a Phase 7 follow-up.
- **No throttling needed.** Delta Chat core already applies a leaky-bucket rate limit to `sendUpdate`; we just call it on every action.
- **No snapshot strategy for MVP.** Only deltas. If history-size becomes a real problem, snapshot+compaction lands as a Phase 7 issue.
- **Multi-device convergence**: payloads carry `addr: webxdc.selfAddr`; only own-`addr` deltas mutate own state. Other peers' updates are leaderboard-only (Phase 6).

## Update payload shape

```js
{ kind: 'delta', addr: <selfAddr>, op: <handlerName>, args: [...], result: {...}, ts: <Date.now()> }
```

## Tasks

- [ ] Define `LocalState` shape in `scripts/state.js` as an **ESM module** (post-#31 toolchain). Mirror MongoDB doc structures from the dd_app handler audit.
- [ ] Implement `applyDelta(state, delta) -> newState` — pure function dispatching to per-op reducers.
- [ ] Boot sequence:
  1. `state = freshState()` from `data/default_game.json`.
  2. `webxdc.setUpdateListener((update) => { state = applyDelta(state, update.payload) }, 0)`.
  3. After replay completes, run materializer (separate issue) once to advance time-based progress.
  4. Engine ready; `Game.js`'s `loadGame` returns this state.
- [ ] Versioning field `state.schema_version` for future migrations.
- [ ] Clock-skew guard: `now = Math.max(Date.now(), state.last_seen_ts)` — wrong system clock never rewinds progress.

## Module rules

- ESM (`export` / `import`), not AMD `define()`.
- No DOM globals (no `window`, `document`, `$`). Importable directly from Node for tests.
- Legacy AMD `Game.js` consumes this via the AMD-compat shim from #31.

## Acceptance

[...]

Issue #9, opened 2026-04-28
Phase 2: integrate stubs, achieve "boots-to-broken" state


experintellia:
Glue PR after the three parallel Phase 2 strip issues.

## Tasks

- [ ] Merge the three Phase 2 branches (auth removal, Socket.js stub, Remote.js+LocalEngine stub).
- [ ] Run the page; it should:
  - Load all assets locally (no external requests).
  - Reach `getToken` and fail with "NotImplemented" — that's the goal.
- [ ] Add `docs/architecture.md` (1 page) describing the post-stub control flow: how a `Game.js` call traverses Remote → LocalEngine and how events flow back via Socket.

## Acceptance

- DevTools Network tab shows zero non-localhost requests.
- DevTools Console shows the expected "NotImplemented: getToken" rejection.
- `docs/architecture.md` committed.

Depends on: all Phase 2 strip issues.

Issue #6, opened 2026-04-28
Phase 2: remove dd_auth integration from bootstrap.js


experintellia:
webxdc provides identity (`webxdc.selfAddr`, `webxdc.selfName`); we don't need `dd_auth`.

## Tasks

- [ ] Remove `#sign_up`, `#sign_in`, `#reset_password`, `#sign_out` hash routes from `scripts/bootstrap.js`.
- [ ] Remove the XHR handlers that POST to `signUpUrl`, `signInUrl`, `resetPasswordUrl`, etc.
- [ ] Remove `setup.js` entries for those URLs (they'll be dead config).
- [ ] Remove `remote.addMethod('checkInvitationToken', ...)` and any auth-only RPC registrations.
- [ ] Boot path should now go straight from page load → `getToken` → `loadGame` (still talking to the soon-to-be-stubbed `Remote.js`, see related issues).
- [ ] Delete `css/dd_auth.css` reference from `index.html` if it only carried auth-page styles.

## Out of scope

- Don't touch `Socket.js` or `Remote.js` — those are separate issues.
- Game can be broken at the end of this — Phase 2 builds up the boot-to-broken state together.

## Acceptance

- `grep -ri "dd_auth\|sign_up\|sign_in\|reset_password" scripts/` returns nothing meaningful.
- Page loads without referencing any auth backend.

experintellia:
Verified complete on master. `bootstrap.js` has no `#sign_up`, `#sign_in`, `#reset_password`, or `#sign_out` routes — only `load` and `downtime`. No `checkInvitationToken`. Boot path goes straight to `getToken` → `loadGame`. All acceptance criteria met. Closing as completed by Wave 1.

---
_Wave 2 orchestrator_

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #273, opened 2026-05-12
fix(render): 10 audit fixes — drag recovery, listener leaks, FX typo, double-remove

experintellia:

## Summary

Three commits covering all 10 audited render-layer bugs. On `master` the render layer is the monolithic `scripts/Render.js`; bugs were located by symptom (typos, `||` coercion, missing listener teardown) and fixed surgically there.

| # | Sev | Symptom | Fix |
| --- | --- | --- | --- |
| 1 | CRITICAL | Drag stuck-down: pointer release outside canvas / alt-tab leaves `dragging=true` forever. | Added `pointercancel`, `blur`, and `visibilitychange` handlers funnelling into `dragend`. New `DragHandler.dispose()` detaches them; `ViewMap.remove` calls it. |
| 2 | HIGH | `config.draggable || true` silently coerced `false` → `true`. | Use `??`. |
| 3 | HIGH | `Sprite.setFrameSrc(src)` early-return used `this.frameSrc` instead of the argument. Src never applied. | Guard on and use `src`. |
| 4 | HIGH | FXFeedMe typo `sacaleY` — animation never played. | Renamed to `scaleY`. |
| 5 | HIGH | FXPuff double-remove from tween onComplete and raw `setTimeout`. | Removal driven only by tween onComplete; raw setTimeout dropped. (FXKatsching's was already commented out.) |
| 6 | HIGH | `Node.remove` left disposed nodes pinned by live tweens. | `Tween.removeTweens(this)` in `remove()`. |
| 7 | HIGH | `dragDelay` setTimeout fired on dead nodes. | `Node.remove` clears `dragDelay` and `cancelClickTimeout`. |
| 8 | HIGH | `Popup.close` rebinds `transitionend` on every call. | `.on(...)` → `.one(...)` (one-shot). |
| 9 | HIGH | `ViewMap` leaks native touch + wheel listeners on teardown (closures pin the whole map). | Stored 4 touch handler refs + wheel ref on the node; new `ViewMap.prototype.remove` detaches them and calls `dragHandler.dispose()`. |
| 10 | HIGH | `SlowTicker` ticks forever even with zero listeners / hidden tab. | Tick short-circuits when listeners empty or `document.hidden`; `addListener` resumes; `visibilitychange` listener restarts on focus. |

## Tests added

`tests/render/render-source.test.js` (11 source-text assertions, one per fix). The render layer can't be instantiated under node-only vitest (needs jQuery / createjs / Scroller globals + real DOM), so these are pure regression guards against the specific source patterns. Runtime behaviour (drag flow, listener teardown, tween race) is covered by the existing Playwright suite.

## Test plan

- [x] `pnpm typecheck` clean.
- [x] `pnpm lint` clean.
- [x] `pnpm test` — 635 passing (was 624; +11 new).
- [ ] Manual drag-recovery smoke test in the dev server (recommended before merge — alt-tab during a drag and confirm the node releases).

## Notes

- The render layer in `claude/plan-code-audit-pt9ul` has been extracted to `scripts/render/*.ts`. When that lands on master, these fixes will need to be re-applied to the corresponding extracted modules (or this PR will need to be rebased on top of the extraction).
- Source-text assertions are intentionally brittle — they're meant to fail loudly the moment someone reintroduces `||` defaults, the `sacaleY` typo, or removes the recovery listeners. Treat them as guard rails rather than functional tests.

---

### DIFF
--- a/scripts/Render.js
+++ b/scripts/Render.js
@@ -350,6 +350,29 @@ var Render = function () {
       e.stopPropagation();
       dm.dragend(e);
     });
+
+    // Audit bug #1: the drag state machine had no recovery paths. If a
+    // pointer was released outside the window, or the user alt-tabbed
+    // away mid-drag, `dragging` stayed `true` forever and every node in
+    // `listeners` kept its `dragging` flag set. The handlers below
+    // funnel pointercancel / window blur / tab-hide visibilitychange
+    // into the same `dragend` path used by mouseup/touchend, and are
+    // stored as refs so dispose() can detach them.
+    dm._recoverHandler = function (e) {
+      dm.dragend(e);
+    };
+    dm._visibilityHandler = function (e) {
+      if (typeof document !== 'undefined' && document && document.hidden) {
+        dm.dragend(e);
+      }
+    };
+    if (typeof window !== 'undefined' && window.addEventListener) {
+      window.addEventListener('pointercancel', dm._recoverHandler);
+      window.addEventListener('blur', dm._recoverHandler);
+    }
+    if (typeof document !== 'undefined' && document && document.addEventListener) {
+      document.addEventListener('visibilitychange', dm._visibilityHandler);
+    }
     this.on('mousemove touchmove', function (e) {
       if (!dm.dragging) {
         return;
@@ -405,6 +428,29 @@ var Render = function () {
           */
       }
     });
+  };
+
+  // Audit bug #1: counterpart to the pointercancel / blur / visibilitychange
+  // listeners installed in init(). Callers that own a DragHandler should
+  // call dispose() before dropping the reference so the global listeners
+  // don't pin the handler.
+  DragHandler.prototype.dispose = function () {
+    if (typeof window !== 'undefined' && window.removeEventListener) {
+      if (this._recoverHandler) {
+        window.removeEventListener('pointercancel', this._recoverHandler);
+        window.removeEventListener('blur', this._recoverHandler);
+      }
+    }
+    if (
+      typeof document !== 'undefined' &&
+      document &&
+      document.removeEventListener &&
+      this._visibilityHandler
+    ) {
+      document.removeEventListener('visibilitychange', this._visibilityHandler);
+    }
+    this._recoverHandler = undefined;
+    this._visibilityHandler = undefined;
   };
 
   ///////////////////////////////////////////
--- a/scripts/Render.js
+++ b/scripts/Render.js
@@ -497,6 +497,16 @@ var Render = function () {
       }
     },
     tick: function () {
+      // Audit bug #10: skip the reschedule when there's nothing to tick
+      // (no listeners) or the document is hidden. `addListener` resumes
+      // it when work shows up, and the visibilitychange listener below
+      // resumes it when the tab becomes visible again.
+      var hidden = typeof document !== 'undefined' && document && document.hidden;
+      var idle = !this.listeners || this.listeners.set.length === 0;
+      this.timeout = undefined;
+      if (hidden || idle) {
+        return;
+      }
       this.listeners.each(function (node) {
         node.tick();
       });
@@ -506,6 +516,10 @@ var Render = function () {
     },
     addListener: function (node) {
       this.listeners.add(node);
+      // Resume ticking if we were idle.
+      if (!this.timeout) {
+        this.start();
+      }
     },
     removeListener: function (node) {
       this.listeners.remove(node);
@@ -516,6 +530,13 @@ var Render = function () {
     },
   };
   SlowTicker.listeners = new Set();
+  if (typeof document !== 'undefined' && document && document.addEventListener) {
+    document.addEventListener('visibilitychange', function () {
+      if (!document.hidden) {
+        SlowTicker.start();
+      }
+    });
+  }
   SlowTicker.start();
 
   /////////////////////////////
@@ -597,6 +618,23 @@ var Render = function () {
     // Remove Node from all references and remove Domobject
     Ticker.removeListener(this);
     SlowTicker.removeListener(this);
+    // Audit bug #6: clear any live tweens targeting this node so a
+    // disposed node is not pinned by an in-flight tween. createjs
+    // Tween is bound at the top of the factory.
+    if (Tween && typeof Tween.removeTweens === 'function') {
+      Tween.removeTweens(this);
+    }
+    // Audit bug #7: dragDelay / cancelClickTimeout setTimeout ids are
+    // never cleared, so a node removed mid-press would still fire its
+    // delayed drag-start or click-cancel handlers against a dead node.
+    if (this.dragDelay) {
+      window.clearTimeout(this.dragDelay);
+      this.dragDelay = undefined;
+    }
+    if (this.cancelClickTimeout) {
+      window.clearTimeout(this.cancelClickTimeout);
+      this.cancelClickTimeout = undefined;
+    }
     if (this.decoratedNode) {
       this.decoratedNode.decorators.remove(this);
     }
@@ -1529,7 +1567,7 @@ var Render = function () {
     //'75%':{'transform':'s1.1'},
     //'100%':{'transform':'s1'}
     //},150);
-    this.FXSimpleCue({ scaleX: 1.1, sacaleY: 1 }, 37);
+    this.FXSimpleCue({ scaleX: 1.1, scaleY: 1 }, 37);
     this.FXSimpleCue({ scaleX: 1, scaleY: 1.1 }, 37);
     this.FXSimpleCue({ scaleX: 1.1, scaleY: 1.1 }, 37);
     this.FXSimpleCue({ scaleX: 1, scaleY: 1 }, 37);
@@ -1557,15 +1595,15 @@ var Render = function () {
   Node.prototype.FXPuff = function (cb) {
     var node = this;
     //node.FXSimple({offsetX:42,offsetY:68,scaleX:1.5,scaleY:1.5,opacity:0},250,'easeOut',cb);
+    // Removal is driven by the Tween onComplete only; the previous
+    // setTimeout fallback (audit bug #5) raced the tween callback and
+    // could double-remove the node.
     node.FXSimple({ scaleX: 1.5, scaleY: 1.5, opacity: 0 }, 250, 'easeOut', function () {
       if (cb) {
         cb();
       }
-    });
-    // make shure it really gets removed (callbacks seem to be unstable)
-    window.setTimeout(function () {
       node.remove();
-    }, 350);
+    });
   };
 
   Node.prototype.FXArise = function (cb) {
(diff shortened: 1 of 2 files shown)


experintellia (commented):
## Review — render-layer audit fixes

**Overview.** Ten bugs in the monolithic `scripts/Render.js`, fixed in three commits: drag-state recovery via `pointercancel` / `blur` / `visibilitychange`, `||` → `??` for `draggable`, `setFrameSrc` argument bug, `sacaleY` typo, FXPuff double-remove, `Tween.removeTweens` in `Node.remove`, drag-timer cleanup, `.on` → `.one` for Popup transitionend, native touch/wheel listener teardown on ViewMap, SlowTicker idle/hidden short-circuit. +313/−98 across 2 files.

### Strengths
- Bug #1 fix shape is correct: refs stored on `dm`, `dispose()` mirrors `init()`, `ViewMap.remove` drives teardown. The `visibilitychange` handler correctly gates on `document.hidden` so it doesn't fire `dragend` on tab-show.
- Bug #2 (`?? true`): exact one-character fix, no ambiguity.
- Bug #3: `setFrameSrc(src)` previously guarded on `this.frameSrc` and never assigned it — the function was effectively dead. The fix also stores `this.frameSrc = src` so subsequent reads work. Clean.
- Bug #6 / #7 in `Node.remove`: `Tween.removeTweens(this)` and clearing `dragDelay` / `cancelClickTimeout` are the standard hygiene for a teardown path.
- Bug #9 (ViewMap listener leak) is the highest-value fix here: anonymous listeners on raw DOM elements were pinning the entire map closure. Moving handlers to instance fields and adding a `remove()` override is the only correct shape; the `Node.prototype.remove.call(this)` tail-call is the right discipline.
- Bug #10 SlowTicker idle/hidden short-circuit: clearing `this.timeout = undefined` *before* the early return is the right ordering, otherwise `addListener` couldn't detect idle state.

### Issues / suggestions
- **Bug #5 (FXPuff) re-introduces a risk the original code was guarding against.** The comment removed by this PR read: `// make sure it really gets removed (callbacks seem to be unstable)`. The `setTimeout` fallback at 350ms was a belt-and-suspenders against the createjs `onComplete` not firing (e.g. when an interrupting tween is started against the same node). Post-fix, if `onComplete` is ever swallowed, the node leaks. Bug #6 mitigates the *other* direction (stale tweens pinning a removed node), but not this one. If the original author's concern was justified, this could surface as occasional FX corpses on the map. Worth a brief test in the actual game (puff a contact, interrupt with another animation, verify removal) before merging.
- **SlowTicker visibilitychange handler may double-schedule.** The new global `document.addEventListener('visibilitychange', …)` calls `SlowTicker.start()` unconditionally on tab-show. If `start()` isn't idempotent (the diff doesn't show its body), a stale `this.timeout` from a tick fired just before the tab hid could combine with the new schedule for two concurrent timers. Worth either adding `if (!this.timeout) this.start();` in the visibility handler or asserting that `start()` already short-circuits.
- **DragHandler recovery handlers attach to `window` per-DragHandler.** If multiple `DragHandler` instances exist (which the `dispose()` pairing implies is possible), each adds its own `pointercancel` / `blur` to window. Probably correct because each handler only acts on its own state, but worth noting that `dragend()` must be a no-op when `dragging === false` — confirmed in the diff context but worth a comment near the recovery handlers.
- **`Tween && typeof Tween.removeTweens === 'function'` guard in `Node.remove`** suggests Tween might be unbound at teardown. If that's a real concern (e.g. unit-test environments where the factory closes over an undefined Tween), the same risk applies to other createjs symbols used in `remove()` — would be cleaner to assert availability once at factory init rather than defensively at every removal.
- **Test coverage is source-text only**, same caveat as #272. The PR description is honest about why (Render.js can't be instantiated under vitest without jQuery / createjs / Scroller globals and a real DOM), and the 11 assertions are well-targeted regression guards. However, the actual behavioral risk lives in *runtime ordering* (drag recovery, FXPuff removal, SlowTicker double-schedule) — none of which is exercised by these tests. The PR plan box mentions "Manual drag-recovery smoke test … recommended before merge" — that should be a blocking checkbox, not an optional one, for the riskiest fix in the batch.
- **Cross-PR collision risk.** PR description correctly notes the render layer is being extracted to TS modules in `claude/plan-code-audit-pt9ul`. If both land, every fix here needs re-applying. Consider landing this PR first (it's smaller and rebases the extraction PR) rather than the other way around.

### Risk
Medium-high among the four. Render layer bugs manifest as UX glitches that don't break tests, and the test suite cannot meaningfully cover behavior.

**Verdict:** LGTM once the manual drag-recovery smoke test is done and FXPuff is sanity-checked under an interrupting animation. The other eight fixes are tight.

---

experintellia:
Thanks for the thorough review — this caught two real issues. Addressed in 30a8a2e.

**1. FXPuff setTimeout regression — fixed properly.** You were right that removing the 350ms fallback re-introduced a leak risk: if an interrupting tween swallows `onComplete`, the node never gets cleaned up. The original double-remove concern is also real (Tween callback + setTimeout could both fire). Reconciled both by:

- Restoring the 350ms `setTimeout(node.remove, 350)` fallback in `FXPuff`.
- Adding an idempotency guard to `Node.prototype.remove` via a `_removed` flag. First call sets `_removed = true`; subsequent calls early-return.

Test updated to assert both halves of the new contract. Now the tween callback racing the setTimeout is safe and a swallowed callback can't leak the node.

**2. SlowTicker visibilitychange double-schedule.** Walked the state machine — `start()` is already idempotent via its own `if (!this.timeout)` check, so the practical race is closed. Added defence-in-depth: the visibilitychange handler now also checks `!SlowTicker.timeout` before calling `start()`. If a future change to `start()` drops its internal guard, the handler still won't double-schedule.

**3. DragHandler dragend idempotency comment.** Added an explicit comment that `dragend` is safe to call multiple times for the same gesture — the recovery handlers (touchcancel + window blur + visibility-hidden) can all funnel for the same release. The empty-listeners loop runs zero iterations on the second call; the `dragging` flag flip is unconditional but harmless when already false. Confirmed in the existing source.

**4. Tween availability defensive check.** Fair point that defensive guards at every removal feel like patches for a real factory-init concern. I'd rather not bikeshed it in this PR — the guard is cheap, the bigger render-extraction work in `claude/plan-code-audit-pt9ul` rewrites this layer end-to-end. Will track as a comment in the extraction PR rather than retrofit it here.

**5. Manual drag-recovery smoke test.** Acknowledged — moving that checkbox to blocking. I can't run a browser session from this environment, so I need to flag this as a pre-merge step that requires human verification. Specifically: alt-tab during a perp drag, confirm the node releases on tab-return.

**6. Cross-PR landing order.** Good call — this PR should land before the render-extraction PR. The extraction would otherwise need to re-derive these 10 fixes from the diff. Will flag this on the extraction PR description.

---

experintellia:
Reviewed ahead of merge.  Fixes are sound; the PR targets stale source.

## Fixes themselves

Spot-checked three of the higher-risk ones:

- **Fix #1 (drag recovery via `pointercancel` + `blur` + `visibilitychange`).**  Correct.  `dragend` is idempotent (`listeners` emptied on first call, `dragging = false` flip harmless when already false), so double-fire from `pointercancel` + `visibilitychange` is safe.  Minor concern: `blur` cancels any in-progress drag on any focus loss — devtools open, popup focus, iframe focus shift.  More aggressive than `pointercancel` alone; worth a comment.
- **Fix #9 (ViewMap native touch + wheel teardown).**  All 4 touch handlers + `_onWheel` are stored on `node` and detached in `ViewMap.prototype.remove`.  Closure-pin claim holds — each handler closes over `node`/`node.scroller`, so stashing a ref on `node` doesn't create *new* pinning.  ✓
- **Fix #10 (`SlowTicker` idle/hidden short-circuit).**  `addListener`'s `if (!this.timeout) this.start()` + `start()`'s own `!this.timeout` guard + the `visibilitychange` restart make the resume path correct.  Calling `addListener` while `document.hidden` re-enters `tick()` which exits early after clearing `this.timeout`, so the ticker stays idle while hidden and resumes cleanly on unhide.  No deadlock.

## Rebase blocker

The render layer has been extracted from `scripts/Render.js` into `scripts/render/*.ts` on master.  The extracted TS modules **still contain every bug this PR fixes** — `sacaleY` typo at `RenderNodeFX.ts:557`, `setFrameSrc` using `this.frameSrc` at `RenderSprite.ts:62`, FXPuff missing the `node.remove()` callback at `RenderNodeFX.ts:537`, the `draggable || true` pattern.  This PR is based on a pre-extraction commit (`146adbc`) and can't cleanly merge to current master.

## Source-text assertion strategy

Defensible for the typo/syntax regressions (`sacaleY`, `config.draggable || true`, `this.frameSrc` vs `src`, `.on(` vs `.one(`) — these are exactly what regex grep can catch and what a careless edit would re-introduce.

**False security for the behavioral fixes:**
- Fix #6 (`Tween.removeTweens(this)` in `remove`): the regex proves the line exists, not that it runs after the new `_removed` guard.  A refactor that reorders the guard would pass the test.
- Fix #1's `DragHandler.dispose()` wiring: the test only asserts `init` mentions the new event names; doesn't verify `dispose()` exists or that `ViewMap.remove` calls it.  A regression that drops the dispose would pass.
- Fix #9: `expect(RENDER_SRC).toMatch(/removeEventListener/)` is near-worthless — the file already had unrelated `removeEventListener` calls.

A JSDOM-backed unit harness for `DragHandler` alone (it only needs `window`/`document`, not jQuery/createjs/Scroller) would cover Fixes #1, #6, #7, #10 with real behavioral assertions.  That single class is cleanly extractable.

## Suggested path

Re-author against `scripts/render/*.ts`.  Rewrite the regex assertions to scan the TS files.  Keep the audit table as the PR description.  Don't merge-and-cherry-pick — the legacy `Render.js` may be vestigial on master and the merge could be reverted, losing the audit history.

---

experintellia:
Closing as obsolete. This PR patched the monolithic `scripts/Render.js`, which was extracted into `scripts/Render.ts` + `scripts/render/*.ts` before this work landed, so the diff no longer applies.

Its entire scope has been re-delivered against the extracted modules across three smaller PRs:

| Original #273 bug | Disposition |
|---|---|
| #2 `draggable \|\| true` → `??` | ✅ #286 (merged) |
| #3 `setFrameSrc` ignores `src` arg | ✅ #286 (merged) |
| #4 `sacaleY` typo | ✅ resolved during the render extraction (no longer present) |
| #5 FXPuff double-remove | ✅ moot — extracted form already correct |
| #6 `Tween.removeTweens` in `remove()` | ✅ #287 (merged) |
| #7 `dragDelay`/`cancelClickTimeout` clear | ✅ #287 (merged) |
| #9 ViewMap native-listener leak | ✅ #287 (merged) |
| #10 SlowTicker idle/hidden short-circuit | ✅ #287 (merged) |
| #1 DragHandler pointercancel/blur/visibility recovery | ✅ #288 (merged) |
| #8 Popup transitionend listener leak | ✅ #288 (merged) |

All ten findings are now in `master`. Nothing here to merge.

---

PR #284, opened 2026-05-13
feat(popups): port notification dialogs to Preact (phase 2 tier 2 of #80)

experintellia:
## Summary

Tier 2 of the phase 2 Preact port (issue #80) — **plus an architecture pivot** away from the `RenderPopup.preactRender` seam introduced in #280.

### Architecture pivot (read this first)

#280 added a `preactRender` field on `RenderPopup` so Preact bodies mounted *inside* the legacy popup container. This PR **deletes that seam** (`RenderTopLevelUI.ts` + `GameNode.openGenericPopup` revert it) and replaces it with a standalone **`scripts/components/popups/dialogManager.ts`** that bypasses `Render.Popup` entirely:

- `openDialog({ component, props, container, extendClass?, placeBottom?, onAfterClose? })` mounts a `<div class="Popup">` Preact tree directly into the active view's `popupContainerDomelem`, manages `.lockOn` / extend-class / centering CSS, and returns a `PreactDialogHandle` (compatible `.trigger / .open / .close / .render` surface so `groot.renderPopup` / `groot.notificationPopup` callers don't change).
- Proper Preact unmount on close (`render(null, container)`), explicit `onClose` injected into every component, no template-vs-Preact branch inside `RenderPopup`.
- **Co-exists with `Render.Popup`** for tier 3+ legacy templates (mission popups, popup_karma, perp popups) — both mount into the same container, mutually exclusive in time.

If you're tracing the #280 `preactRender` entry point, it's gone — the entry point is now `GameRoot.openPreactDialog` → `dialogManager.openDialog`.

## What ships (tier 2 dialogs)

* **`LevelUpNotification.tsx`** — replaces `views/levelup.html`. xp_level snapshotted at cue time so the component is pure.
* **`NewItemsNotification.tsx`** — replaces `views/notification.html` + the `notification_item.html` partial (now an internal subcomponent). Drives `data.perps` + `data.powerups`. Cue projects only the four perp fields the component reads (no full `type_data` pin).
* **`TutorialNotification.tsx`** — replaces `views/notification_tutorial.html`. Drives `data.simplemessage` / `data.story` / `data.tutorial`. Tap-anywhere dismiss via the component's own `onClick={onClose}`.

Deleted: `views/levelup.html`, `views/notification.html`, `views/notification_item.html`, `views/notification_tutorial.html`.

## Scope decision: Karma incident deferred to tier 6

Karma incident shares `popup_karma.html` with the Karma selector status popup (tier 6). Porting one half now would render the same template via two systems. Both karma flows port together in a tier-6 PR.

## Review findings — status

| # | Finding | Status |
|---|---|---|
| placeBottom regression (blocker) | Preact dialogs ignored `placeBottom` — and in fact were mispositioned to **top-left** entirely (no RenderNode transform). | ✅ Fixed — `.PopupContainer.lockOn.PopupPreact` flex-centers; `.PopupPreactBottom` bottom-docks. Verified centered at viewport center. |
| Bug 1 — backdrop listener leak | (was on pre-/simplify code) | ✅ Already fixed — `close()` removes its own listeners |
| Bug 2 — scriptedEvents not cancel-guarded | story camera-pan ran for a dismissed cue | ✅ Fixed — `delayScript` timeout now checks `isOpen` |
| Bug 3 — stale jQuery-handler comments | Level/Tutorial headers cited deleted seam | ✅ Fixed |
| Bug 4 — no-op `open` setter | (was on pre-/simplify code) | ✅ Already fixed — interface is `readonly open`, no setter |
| Close animation lost | synchronous `render(null)` skips the 0.2s fade | ⏳ **Deferred** — naive fix introduces an open-over-open clobber hazard (deferred unmount vs. the next dialog mounting into the same container). Tracked for a focused follow-up. |
| Type-erased cue→component dispatch | `ComponentType<any>` + cast | ⏳ **Deferred to tier 3** — a discriminated-union dispatcher pays off when tier 3 doubles the component count; premature for 3 components. |
| `popup_cancel` not aliased on X | dormant (no listener on Preact dialogs); manager's `trigger()` already aliases both | ⏳ Deferred — no functional need until a listener exists |
| Manager handler surface (tab strip / pagination / etc.) | tier 4+ direction | 📋 Planning note — tier-4 components own their own handlers (validated against phase-1 Section I selectors); decision recorded for the tier-4 briefing |

## Validation

- [x] `pnpm test` — 667 unit
- [x] `pnpm exec playwright test` — 91 passed, 1 pre-existing skip; all 42 phase-1 dialog-lifecycle tests green
- [x] Position verified: Cash popup centers at viewport center; placeBottom cues bottom-dock
- [ ] Reviewer smoke: level-up (gain XP), new-items (unlock a perp on level-up), tutorial/story sequence, **and confirm dialogs are centered + bottom-docked variants look right**

## Progress (issue #80)

| Tier | Dialogs | Status |
|---|---|---|
| 1 — status info | Profiles, Cash, AP, XP | ✅ #280 |
| 2 — notifications | Level-up, New items, Tutorial/story | **this PR** |
| 2 — Karma incident | shares template with tier 6 | grouped with Karma selector |
| 3 — missions | Briefing, Complete | next |
| 4-8 | settings, perps, catalogs, cleanup | later |

7 of 22 dialogs ported.

---


experintellia:
Two review passes ran on this branch ahead of merge.  High-level findings below.

## Architecture pivot worth calling out

This PR doesn't extend the `RenderPopup.preactRender` seam from #280 — it deletes it (the field is reverted in `RenderTopLevelUI.ts` and `GameNode.ts`) and replaces it with a parallel `dialogManager` (`scripts/components/popups/dialogManager.ts`) that bypasses `Render.Popup` entirely.  The PR body says "mirrors the option added to `PopupConfig` in #280", which understates the change.  The pivot is defensible — proper Preact unmount on close, cleaner component ownership, no template-vs-Preact branch inside `RenderPopup` — but worth saying plainly in the description so future readers don't trace the deleted seam looking for the entry point.

## Concrete issues

**1. `placeBottom` regression (blocker).**  Every Tutorial / Story / LevelUp cue still sets `config.placeBottom: true` (`GameRoot.ts:1520`, `1700`, `1714`, …).  The legacy `RenderPopup` reads this to dock the popup at the bottom of the viewport.  `dialogManager` never reads it.  LevelUp + Story + Simplemessage popups will render in their default position instead of bottom-docked.  The phase-1 test contract passes green because Section B asserts only sprite + extendClass + open/close, not position — exactly the under-specification the original spec docstring warned about ("What this spec does NOT verify: Visual appearance, mobile layout").

**2. Close animation lost.**  Legacy `Popup.close()` defers `remove()` until after the 250–500ms CSS transition.  `dialogManager.close()` calls `render(null, container)` synchronously, so the popup disappears instantly — the fade-out is gone.  Tests still pass (3s timeout absorbs it).  Re-add by toggling a `.closing` class and calling `render(null, …)` from a `setTimeout` matched to the existing transition duration.

**3. Cue → component dispatch is type-erased.**  `NotificationConfig.component: ComponentType<any>` plus `props: Record<string, unknown>` (`GameRoot.ts:316-326`), then cast back at `openPreactDialog` to `ComponentType<Record<string, unknown> & { onClose: () => void }>`.  A future rename of `LevelUpNotificationProps.xpToNext` won't surface a compile error at `makeNotifications` — the next refactor breaks the level-up popup silently.  Worth a typed dispatcher (discriminated union of `{ variant: 'levelup', props: LevelUpNotificationProps } | { variant: 'newItems', props: NewItemsNotificationProps } | …`) before tier 3 doubles the number of components.

**4. `PopupClose` X fires only `popup_close`, not `popup_cancel`.**  Legacy delegated handler at `RenderTopLevelUI.ts:663-668` fires both.  No caller in the codebase listens for `popup_cancel` on a Preact-managed dialog, so this is dormant — but worth harmonizing in the manager's `trigger()` so the two events stay aliases everywhere.

**5. Forward-looking — `dialogManager` re-implements only subpop close + FX feedback.**  Legacy `initBaseUI` also installs delegated handlers for `.PopupMenuButton` (tab strip), `.Powerup`, `.PopupPerp`, `.PopupToken`, `.Pagination` arrows, `extendClass === 'Tutorial'` tap-anywhere advance, `mousemove`/`mousedown` position tracking, and the `close_powerup` event.  Tier 4 settings (multi-tab) and tier 5+ powerup/perp/token popups will need either component-internal handlers (clean for new code) or expanded manager surface — pick before tier 4 lands to avoid two competing patterns.  Phase-1 Section I (tab strip) asserts on `.PopupMenuButton[data-tab=…]` + `.active` flip, so a component-only port must reproduce both the selectors *and* the class-flip behavior.

## What works as advertised

- Preact unmount on close (`dialogManager.ts:84`) — closes the gap that was open after #280.
- DOM equivalence with deleted templates (LevelUp, NewItems, Tutorial) — verified element by element.
- Backdrop close matches the legacy "any popup without `.NoClose`" behavior at `RenderTopLevelUI.ts:586-594`.  ✓
- `notificationPopup` stub-then-live race during the `delay` window: the `cancelled` flag guards correctly against the setTimeout firing after a pre-delay `trigger('popup_close')`.  ✓
- FX feedback (`no_cash` / `no_AP` / `error`): `handle.lastButton` structural-compat preserves the Section F test path verbatim.  ✓

## Suggested merge condition

Fix `placeBottom`.  The animation regression and the type-erasure are reasonable follow-ups.  Document the pivot from the #280 seam in the PR description, and pick the manager-handler-surface direction in the phase-2 briefing before tier 4 starts.

---

experintellia:
## Code-level correctness pass (follow-up to the architecture comment above)

Line-level review of the new code in `dialogManager.ts` + the `GameRoot` wiring — the newest/riskiest surface.

### Bug 1 — backdrop dismissal leaks two listeners on a long-lived node (should fix, ~2 lines)

The container handler is registered with two close paths:

```ts
opts.container.addEventListener('click', handleInteraction);
opts.container.addEventListener('touchend', handleInteraction);

const cleanup = close;
const wrappedClose = (): void => {
  opts.container.removeEventListener('click', handleInteraction);
  opts.container.removeEventListener('touchend', handleInteraction);
  cleanup();
};
handle.close = wrappedClose;
```

Only `wrappedClose` removes the listeners.  But the backdrop branch inside `handleInteraction` calls the **un-wrapped** `close()`:

```ts
const handleInteraction = (e: Event): void => {
  if (e.target === opts.container) {
    close();          // ← listeners never removed
    return;
  }
```

`opts.container` is `groot.renderNode.popupContainerDomelem[0]` — a **persistent** DOM node reused for every popup for the life of the view, not recreated per dialog.  So every backdrop dismissal (and every Tutorial / LevelUp / Story tap on the backdrop, since those close through this same generic backdrop path) leaves the old `click` + `touchend` listeners attached.  Open the next popup → a second `handleInteraction` is added.  Clicks then fire **both**; the stale one's `close()` early-returns on `!isOpen`, but its subpop branch still runs `e.stopPropagation()` / `preventDefault()` plus the jQuery DOM mutation against whatever is currently mounted.  Accumulating leak + multiple-fire hazard over a session.

Close paths audited:
- X button / MainButton → `onClose` prop = `wrappedClose` — clean
- `handle.trigger('popup_close')` → reassigned to `wrappedClose` — clean
- `active.handle.close()` (open-over-open) → `wrappedClose` — clean
- **backdrop / tutorial-backdrop tap → `handleInteraction` → `close()` — leaks**

Fix: the backdrop branch should call `wrappedClose()`.  It's declared after `handleInteraction`; resolve by hoisting the close logic or calling through `handle.close()` (which is `wrappedClose` by the time any event can fire).

### Bug 2 — `scriptedEvents` fire even when the cue was cancelled during the delay window (edge case)

`GameRoot.openNotification`, Preact branch:

```ts
window.setTimeout(() => {
  notification.scriptedEvents?.forEach((s) => s());
}, config.delayScript ?? 0);

window.setTimeout(() => {
  if (cancelled) return;      // ← delay path guarded
  ...
}, config.delay ?? 0);
```

The `delay` path checks `cancelled`; the `delayScript` path does not.  If the player dismisses (stub `trigger('popup_close')` sets `cancelled = true`) before `delayScript` elapses, the scripted events still run.  For story tutorials those pan the camera to `storyPerp.ViewMap` — the camera jumps to a perp for a notification the player already dismissed.  Add the same `if (cancelled) return;` guard to the `delayScript` timeout.

### Bug 3 — stale comments describe a mechanism that no longer applies (doc correctness)

`NewItemsNotification.tsx` header: "Dismissed via the OK MainButton — the jQuery delegated handler in `RenderTopLevelUI.ts` fires `button_click.MainButton`, which the popup-events init in `GameNode.initPopupEvents` translates into `popup_close`."  That path doesn't exist for this component — `dialogManager` bypasses `Render.Popup`; the MainButton dismisses via its own `onClick={close}`.  Same issue in `LevelUpNotification.tsx` / `TutorialNotification.tsx`, which cite the `RenderTopLevelUI.ts` `extendClass === 'Tutorial'` jQuery handler; actual dismissal is the component's own `onClick={onClose}` on `.TutorialBody`.  These will send a future debugger to the wrong file.

### Bug 4 — no-op `open` setter silently discards writes (minor)

```ts
get open() { return isOpen; },
set open(_v: boolean) { /* read-only via getter */ },
```

Legacy `RenderPopup.open` is a writable flag.  A legacy caller doing `popup.open = false` gets silent no-op semantics rather than the expected state change.  No such caller exists in the current tree (phase-1 spec drains via `trigger('popup_close')`), so it's latent — but a silent-discard setter is worse than none for catching a future regression.  Either make it set `isOpen`, or drop the setter and let a stray write throw.

### Verified correct (not bugs)

- Open-over-open (`if (active) active.handle.close()`) runs the previous `wrappedClose`; listeners removed, `active` reset before the new mount.
- `notificationPopup` stub→live handoff: synchronous gap between `delete` and reassign, no interleaving; `nonblocking` auto-close is idempotent via `close`'s `if (!isOpen) return`.
- Double `touchend` + synthetic-`click` on backdrop: second invocation early-returns on `!isOpen` (functionally safe; the leak is Bug 1, a separate root cause).
- DOM structure (`.PopupContainer.lockOn > .Popup > .PopupBody …`) matches legacy CSS selectors and the phase-1 test contract.

---

Bug 1 is the one I'd block on alongside the `placeBottom` regression from the previous comment — a real accumulating leak with a multi-fire hazard, and a ~2-line fix.  Bugs 2–4 are reasonable follow-ups.

---

experintellia:
Addressed the review in `4cf42d8`. Summary:

**Blocker fixed — positioning.** The `placeBottom` finding undersold it: I measured the Cash popup at `x:6 y:54` on a 1280×720 viewport — *every* Preact dialog was pinned top-left, because the legacy centering was the RenderNode `setTransform` translate3d, which the manager doesn't have. Fixed with `.PopupContainer.lockOn.PopupPreact` flex-centering + `.PopupPreactBottom` for the legacy `placeBottom` cues (Tutorial/Story/LevelUp), toggled by the manager on open/close and threaded through `openPreactDialog` → `openNotification`. Verified: Cash now centers at viewport center; placeBottom cues bottom-dock.

**Bug 2 fixed.** `delayScript` timeout now checks `isOpen` — a story tutorial dismissed during its delay window no longer runs its scripted camera-pan.

**Bug 3 fixed.** `LevelUpNotification` / `TutorialNotification` headers + the inline tutorial comment no longer cite the deleted `RenderTopLevelUI` jQuery seam.

**Bug 1 & Bug 4 were on pre-/simplify code** — verified already fixed in the current tree: `close()` removes its own `click`/`touchend` listeners (no backdrop leak), and `PreactDialogHandle.open` is `readonly` with no setter.

**Deferred (you flagged these as follow-ups):**
- **Close animation** — the naive re-add (defer `render(null)` behind the 0.2s `.close` transition) introduces an open-over-open clobber: the deferred unmount fires `render(null, container)` *after* a displacing dialog has mounted into the same container, wiping it. Doing it safely needs a generation guard + a synchronous-teardown path for the displacement case. Worth a focused follow-up rather than rushing it into this PR.
- **Type-erased dispatch** — agree a discriminated-union dispatcher is the right shape; it pays for itself once tier 3 doubles the component count. I'll set it up at the start of tier 3 rather than carry the abstraction for 3 components.
- **`popup_cancel` alias** — the manager's `trigger()` already treats `popup_close`/`popup_cancel` as aliases; the only gap is the X button calling `close()` directly rather than routing through `trigger('popup_cancel')`, and nothing in the tree listens for `popup_cancel` on a Preact dialog. Will harmonize when a listener actually exists.
- **Manager handler surface for tier 4+** — recording the decision now: tier-4 settings (multi-tab) and tier-5+ perp/token popups own their handlers *component-internally* (clean for new code), reproducing the phase-1 Section I selectors (`.PopupMenuButton[data-tab=…]` + `.active` flip) inside the component rather than expanding the manager's delegated surface. This goes in the tier-4 briefing so we don't end up with two competing patterns.

CI green; 42 phase-1 dialog tests + 667 unit pass. The PR description now documents the #280→dialogManager pivot up front.

---

PR #285, opened 2026-05-13
fix(engine): level up when xp lands exactly on xp_max

experintellia:
## Summary

Reported during manual testing of #272: in some cases the level-up notification doesn't fire even though the XP bar shows "10/10", and as a downstream symptom the **Bogus Company** (`proxy001`, `required_level: 2`) buy fails in the early game.

## Root cause

`_getLevelByXP` in `scripts/LocalEngine.ts:1117` matched levels with an *inclusive* upper bound:

```ts
if (lvl && xp >= lvl.xp_min && xp <= lvl.xp_max) return lvl.number;
```

Ruleset levels have a 1-XP gap between consecutive ranges:
- L1: `xp_min=0, xp_max=10`
- L2: `xp_min=11, xp_max=30`

At `xp=10`, the loop matches L1 first (`10 <= 10`) and returns. `_checkLevelup(currentLevel=1, newXp=10)` is then `1 > 1` = **false** → no level-up. The player is stuck on level 1 with "10/10" until the next XP event pushes them past 11.

**Chain reaction:** since `xp_level` stays at 1, `buyPerp('proxy001')` fails its `required_level=2` check (`LocalEngine.ts:1403`) and returns `error: 1`. The Bogus Company purchase silently errors out in the early game even though the UX implies it should work after the first mission completes.

**Why it's "sometimes":** only matters when an XP reward lands the player *exactly* on `xp_max` (e.g. xp=8 + a mission reward of 2 → xp=10). Players whose increments overshoot to 11+ level up normally.

## Fix

Walk levels high-to-low and treat `xp_max` as the promotion threshold:

```ts
for (var i = levels.length - 1; i >= 0; i--) {
  var lvl = levels[i];
  if (lvl && xp >= lvl.xp_min) {
    var next = levels[i + 1];
    if (next && xp >= lvl.xp_max) return next.number;
    return lvl.number;
  }
}
```

No ruleset data change. Players whose `xp_value` lands exactly on a `xp_max` now promote immediately. The matching visual check in `GameRoot.setXP` switches from `>` to `>=` so the status-bar level setter agrees with the engine.

## Test plan

- [x] New `tests/handlers/levelup-xp-max.test.js` directly exercises the regression — starts at xp=9 / level=1, buys client007 (`xp_inc=1`) to land at exactly 10, asserts `xp_level === 2` and the levelup achievement fires once.
- [x] Verified the new test **fails** without the fix (stashed the LocalEngine + GameRoot changes locally and saw the failure) — and passes with it.
- [x] Counter-test: xp=5 → xp=6 does **not** trigger levelup.
- [x] `pnpm typecheck` clean.
- [x] `pnpm test` — 680 passing (was 678 + 2 new).
- [x] `pnpm lint` clean.

## Related

- Originally listed as MEDIUM finding in the multi-agent code audit (`GameRoot.ts:1932 "XP until next level" off-by-one when at level cap`). Didn't make it into the 4 audit PRs (#270–#273); pulled out as a focused fix when a real player hit it.
- Closes the manual-test regression report from #272's post-merge sanity check.

https://claude.ai/code/session_01NhYApbEtp5AYkzbJyD5mnf

---

### DIFF
--- a/scripts/LocalEngine.ts
+++ b/scripts/LocalEngine.ts
@@ -1115,13 +1115,27 @@ function _tokensEqual(a: TokenSpec[], b: TokenSpec[]): boolean {
 }
 
 function _getLevelByXP(xp: number): number {
+  // Ruleset levels are defined with a gap between consecutive ranges
+  // (L1: 0–10, L2: 11–30, …). A player whose xp lands exactly on a
+  // level's `xp_max` would otherwise stay on the lower level until the
+  // next +1 XP pushed them past the gap into the next `xp_min`. That's
+  // visible as "10/10" sitting at level 1 with no levelup notification.
+  //
+  // Treat `xp_max` as the promotion threshold: walk high-to-low and
+  // return the first level whose `xp_min` is reached, but if xp has
+  // already met or passed that level's `xp_max` AND a higher level
+  // exists, return the higher level instead.
   var levels = _getRuleset().levels;
-  for (var i = 0; i < levels.length; i++) {
+  for (var i = levels.length - 1; i >= 0; i--) {
     var lvl = levels[i];
-    if (lvl && xp >= lvl.xp_min && xp <= lvl.xp_max) return lvl.number;
+    if (lvl && xp >= lvl.xp_min) {
+      var next = levels[i + 1];
+      if (next && xp >= lvl.xp_max) return next.number;
+      return lvl.number;
+    }
   }
-  var last = levels[levels.length - 1];
-  return last ? last.number : 1;
+  var first = levels[0];
+  return first ? first.number : 1;
 }
 
 function _checkLevelup(currentLevel: number, newXp: number): boolean {
--- a/scripts/game/GameRoot.ts
+++ b/scripts/game/GameRoot.ts
@@ -1178,7 +1178,7 @@ export class GameRoot extends GameNode {
     if (num !== undefined && num > this.xp_value) {
       this.xp_value = num;
     }
-    if (this.xp_value > this.xp_level.xp_max) {
+    if (this.xp_value >= this.xp_level.xp_max) {
       this.setLevel(this.getLevelByXP(this.xp_value).number);
     }
     if (this.xp_value < this.xp_level.xp_min) {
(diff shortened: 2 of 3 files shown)


experintellia:
Reviewed ahead of merge.  The engine fix is correct, but two adjacent code paths weren't updated and one of them silently undoes the sibling fix in this PR.

## Engine `_getLevelByXP` rewrite (`LocalEngine.ts:1109-1138`)

Correct.  Handles level cap (`next` is `undefined` → returns top level), xp=0, and the `[xp_min, xp_max)` window cleanly.  Works for non-1-XP gaps too — verified across all 60 levels in `data/ruleset_3.{en,de}.json`.

## Unfixed duplicates

**1. `GameRoot.getLevelByXP` at `GameRoot.ts:980`** still uses the old inclusive predicate:

```ts
(this.data.levels ?? []).find((lvl) => xp >= lvl.xp_min && xp <= lvl.xp_max)
```

This is what `setXP` calls *after* the new `>=` trigger fires.  At `xp == xp_max`, it returns the *lower* level entry, so `setLevel(getLevelByXP(xp).number)` is a no-op and the status bar stays at "10/10 — Level 1" until xp hits 11.  **The status-bar `>` → `>=` fix in this PR is silently undone by this duplicate.**

**2. `LocalEngine.ts:858`** (`_findLevelByXP` inside `buyKarma`) has the same off-by-one and wasn't touched.  Karma purchases landing xp exactly on `xp_max` will compute AP off the *lower* level entry.

## Test coverage

`tests/handlers/levelup-xp-max.test.js` pins the regression well, but:
- No level-cap boundary test — should assert that landing xp exactly on `L60.xp_max` does *not* fire a levelup achievement (no L61).
- No multi-step promotion test — e.g. xp jumps from L1's range straight past L2's window into L3.
- Comments in `tests/handlers/collect-integrate.test.js:843` and `tests/handlers/achievements.test.js:193` describe the *old* boundary semantics ("client007 has xp_inc=1, so xp_value 10→11 → level 2").  They still pass with the new logic because they cross the gap entirely, but the stale comments could invite a future "restoration" of the old behavior.

## Suggested merge condition

Apply the same fix at `GameRoot.ts:980` and `LocalEngine.ts:858`.  Add the cap-boundary test.  Then good to merge.

---

PR #289, opened 2026-05-15
fix(game): restore mission tab for players who finished all missions (regression from #272)

experintellia:
## Summary

**Regression fix.** After #272 merged, the mission tab renders **empty for any player who has completed every available mission**.

## Root cause

#272 removed `Missions.initMissions`' `else` branch — which marked *every* loaded mission `complete=true` whenever `active_missions` was empty — because it re-fired spurious `mission_complete` events on every reload. Removing it was correct for the spurious-event bug, but introduced a worse one:

When a player finishes the last available mission, the engine producer (`LocalEngine.ts`) emits `active_missions: []`. On reload, `initMissions` runs with an empty list, so the `active_missions` walk sets neither `active` nor `complete` on any mission. `getVisibleMissions()` filters on `active === true || complete === true` → returns `[]` → **empty mission tab**.

The #272 reviewer flagged this exact risk ("if any production path *does* send `active_missions: []` to mean 'all done'... the player will no longer see missions flip to complete on reload"). It was incorrectly dismissed at the time — owning that here.

## Fix

After the `active_missions` walk, independently derive each mission's `complete` flag from goal state: a mission whose goals are all complete is marked complete. This:

- Restores the finished-mission history in the tab when `active_missions` is empty.
- Is the correct source of truth (matches the engine-side `_completeMissionsIfReady` logic).
- Unlike the old blanket `else`, never marks a never-started mission complete — so the original #272 spurious-event bug stays fixed.

## Test plan

- [x] `tests/game/missions-init.test.js` extended: keeps the original `Array.isArray` guard assertions, adds two pinning the goal-derived pass and verifying it runs *outside* the `active_missions` length guard.
- [x] Confirmed the two new assertions **fail** without the source fix (stashed it, saw 2 failures), pass with it.
- [x] `pnpm typecheck` clean.
- [x] `pnpm test` — 696 passing.
- [x] `pnpm lint` clean.
- [x] `pnpm build:all` clean.

## Manual verification recommended

Load a save where every mission is finished (or play through to all-complete, then reload). The mission tab should show the completed missions, not be blank.

## Commits

1. **fix(game): restore mission tab for players who finished all missions** — the fix + regression tests.
2. **chore(game): simplify per simplify-skill review** — use canonical `MissionGoal` type, tighten naming/comment. No behavior change.

https://claude.ai/code/session_01NhYApbEtp5AYkzbJyD5mnf

---

### DIFF
--- a/scripts/game/Missions.ts
+++ b/scripts/game/Missions.ts
@@ -143,7 +143,11 @@ export class Missions extends GameNode {
       allGoalsDone.set(g.mission, prev === undefined ? done : prev && done);
     }
     for (const mission of Object.values(this.Missions)) {
-      if (mission.gestalt && allGoalsDone.get(mission.gestalt) === true && !mission.states.complete) {
+      if (
+        mission.gestalt &&
+        allGoalsDone.get(mission.gestalt) === true &&
+        !mission.states.complete
+      ) {
         mission.setState('complete', true);
         mission.setState('active', false);
       }
--- a/scripts/game/Missions.ts
+++ b/scripts/game/Missions.ts
@@ -109,13 +109,12 @@ export class Missions extends GameNode {
     if (raw_data.mission_goals) {
       this.updateMissionGoals(raw_data.mission_goals);
     }
-    // Only walk active_missions when the server actually handed us a list.
-    // Pre-fix, a falsy/empty `active_missions` (undefined or []) hit an
-    // unguarded else-branch that flagged *every* loaded mission complete,
-    // which on every reload re-fired spurious mission_complete state
-    // events for missions the player had not finished. The server uses
-    // a non-empty list of completed-mission gestalts to encode "all
-    // done"; an empty list simply means "nothing to mark active here".
+    // Walk active_missions when the server handed us a non-empty list:
+    // mark each active and its required-mission ancestors complete.
+    // Pre-fix, a falsy/empty `active_missions` hit an unguarded
+    // else-branch that flagged *every* loaded mission complete (even
+    // ones never started), re-firing spurious mission_complete events
+    // on every reload.
     if (Array.isArray(raw_data.active_missions) && active_missions.length) {
       active_missions.forEach((gestalt) => {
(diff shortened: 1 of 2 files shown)


experintellia:
## Review — looks correct, safe to merge

Traced the full path: engine producer → `initMissions` → `setState`→event → #272 history.

**Correctness verified against the engine.** The goal-derived fold (`prev === undefined ? done : prev && done`) is a precise mirror of `LocalEngine._completeMissionsIfReady` (`scripts/LocalEngine.ts:1844-1860`), which completes a mission via `goals.every(g => g.complete)` and filters it out of `active_missions` while keeping the completed goal rows in `mission_goals`. Edge cases line up: no goal rows → not completed (matches engine's `if (!goals.length) return true`); partial → not completed; `mission_goals` rows only exist post-activation, so "all rows complete" reliably means finished.

**No spurious-event regression.** The stronger reason than "never touches a never-started mission": the post-#272 active-walk *already* does the structurally identical `m.setState('complete', true); m.setState('active', false)` for `getBranch()` ancestors on every reload (accepted behavior). The new pass is equivalent; the `!mission.states.complete` guard + `setState`'s no-op-on-same-value early return (`GameNode.ts:456`) prevent double-firing. No new event class introduced.

**Minor caveats:**
1. The "goal-derived pass is outside the length guard" test relies on `indexOf('\n    }')` hitting the active-walk's 4-space closing brace — works now, but commit 7f131e4 was itself a biome reformat of this region, so a future reformat could silently move the match. Acceptable given the documented source-text-guard convention; flagging as a known fragility.
2. Worth actually doing the suggested manual check (all-complete save, reload x2): confirm the tab populates **and** no "Mission complete!" popup/`FXMissionComplete` re-fires. The structural-equivalence argument says it's fine, but `renderNode` attachment timing during `loadGame`→`initMissions` is the one thing not fully pinnable from source.

Types/style clean (canonical `MissionGoal`, necessary cast, defensive `typeof` guard, two negligible O(n) passes). Risk: low.

---

PR #291, opened 2026-05-16
feat(popups): port mission dialogs to Preact + typed dispatcher (phase 2 tier 3 of #80)

experintellia:
## Summary

Tier 3 of the phase-2 Preact port (issue #80): mission **briefing** + mission **complete** dialogs, plus the **typed cue→component dispatcher** that was the deferred follow-up from the #284 review.

## Typed dispatcher (deferred from #284)

`scripts/components/popups/dialogRegistry.ts` — a `DialogSpec` discriminated union maps each notification-cue variant to its component's props type, resolved via `resolveDialog`. `NotificationConfig`'s loose `component: ComponentType<any>` + `props: Record<string, unknown>` is replaced by a single typed `config.dialog`. A prop rename (e.g. `LevelUpNotificationProps.xpToNext`) now surfaces a **compile error at the cue site** instead of silently breaking the popup at runtime. The existing tier-2 cues (levelup / newItems / tutorial) are migrated onto it.

## Mission dialogs

* **`MissionPopup.tsx`** — one component, `briefing` | `complete` variant (`.PopupBody.MissionBody` / `.MissionBody.Complete`). Replaces `views/popup_mission.html` + `views/popup_mission_complete.html`.
* **`missionView.ts`** — `buildMissionPopupProps` ports the per-goal workflow text / sprite / progress computation out of the `mission_goal.html` template (which literally carried a `MOVE THIS TO GAME` FIXME) and the `mission_rewards.html` classMap. Feeds both the notification cue path and `Mission.openMissionPopup` (direct mission-card click).
* **Briefing dismissal persistence** — `dismissMissionBriefing` (stops the briefing re-queuing on webxdc replay) was wired via `popup.notificationMission` → `GameNode.initPopupEvents` popup_close. The Preact path has no `initPopupEvents`, so it now fires from the cue's `drainOnClose`, gated on `mission_active_gestalt`. Same persisted-op semantics.

## Template cleanup — partial scope

Deleted: `views/popup_mission.html`, `views/popup_mission_complete.html`.

**Kept:** `views/mission_goal.html`, `views/mission_rewards.html` — `views/mission.html` (the not-yet-ported mission-perp **card** in the Missions tab) still renders them. The goal/reward logic is therefore duplicated between `missionView.ts` and those partials during the transition (same pattern as the SubpopClose duplication in the dialog manager). Both collapse in tier 8 when the mission card ports. (Caught via a "view not bundled" console warning after an initial over-delete — now covered.)

Also dropped the dead `Mission.popupTemplate` field (Mission extends GameNode, not GamePerp — never read).

## Validation

- [x] `pnpm test` — 704 unit
- [x] `pnpm exec playwright test tests/e2e/dialog-lifecycle.spec.ts` — 43 passed (incl. Section B "Mission briefing" / "Mission complete"); no "view not bundled" warnings
- [ ] Reviewer smoke: open a mission card (briefing — goals/rewards/progress render, "Close" vs "mission button" by active state), complete a mission (the `.Complete` popup), confirm a dismissed briefing does **not** re-appear after reload (persistence)

## Progress (issue #80)

| Tier | Dialogs | Status |
|---|---|---|
| 1 — status info | Profiles, Cash, AP, XP | ✅ #280 |
| 2 — notifications | Level-up, New items, Tutorial/story | ✅ #284 (+ #290 level-up HTML fix) |
| 3 — missions | Briefing, Complete | **this PR** |
| 4 — settings | User data | next |
| 5-8 | perps, catalogs, cleanup | later |

9 of 22 dialogs ported.

---


experintellia:
## Review — tier 3 (mission dialogs + typed dispatcher)

First, the #284 follow-ups: confirmed in the merged `dialogManager.ts` that the listener-leak and `open`-setter findings are genuinely resolved — single `close()` that removes its own listeners, `readonly open`. The "was pre-simplify code" note checks out.

### Typed dispatcher — delivers exactly the deferred #284 follow-up (positive)

`dialogRegistry.ts` is done right.  `DialogSpec` is a discriminated union mapping each cue variant to `Omit<ComponentProps, 'onClose'>`; a rename of e.g. `LevelUpNotificationProps.xpToNext` now produces a **compile error at the cue site** in `makeNotifications` instead of a silent runtime break.  The `Omit`-at-cue / inject-`onClose`-at-manager split is consistent end to end.  Direct opens (status popups, `Mission.openMissionPopup`) correctly stay on the generically-typed `openPreactDialog`, which still type-checks props.  The loose `ComponentType<any>` + `Record<string, unknown>` is gone.

### DOM equivalence — verified faithful (positive)

`MissionPopup.tsx` matches the deleted `popup_mission.html` / `popup_mission_complete.html` at the top level, and the per-goal / per-reward DOM matches the **kept** `mission_goal.html` / `mission_rewards.html` element-for-element — including legacy quirks reproduced exactly: the hardcoded `<div class="RenderSprite Tobi">` and the unformatted `print(reward.amount)`.  `buildMissionPopupProps` reproduces the deleted templates' variant logic (`.Complete` modifier; `mission button` vs `Close` vs `mission_done button`; `mission_done text` vs `data.description`).  `buildGoal`'s workflow-text branching is a line-accurate port of `mission_goal.html`'s switch.  One trivial *positive* deviation: an unmapped reward target now yields `class="MissionReward PopupSummaryItem "` vs the legacy literal `undefined` — cosmetically identical, marginally cleaner.

### Findings

**1. Dual-discriminant footgun (real, minor).**  `DialogSpec.variant` (`missionBriefing`/`missionComplete`) and `MissionPopupProps.variant` (`briefing`/`complete`) are independent values the cue site must hand-sync:

```ts
dialog: { variant: 'missionComplete', props: buildMissionPopupProps({ ..., variant: 'complete' }) }
```

`variant: 'missionComplete'` paired with `buildMissionPopupProps({ variant: 'briefing' })` compiles cleanly and renders the wrong body modifier.  The registry maps both mission variants to the same component without constraining `props.variant`.  Worth either deriving the component variant from the spec, or collapsing to a single discriminant.

**2. Live logic duplication with an explicit multi-tier lifetime (maintenance hazard — PR acknowledges it).**  `missionView.ts:buildGoal` + the reward classMap now duplicate the kept `mission_goal.html` / `mission_rewards.html`, which `views/mission.html` (the not-yet-ported mission card) still renders.  Verified they're *currently* equivalent, but they must stay in sync until tier 8.  A ruleset/text change touching one path and not the other → the popup and the mission card show different goal text while sharing the same CSS.  Given the duplication is planned to live ~5 tiers, recommend a test that renders one fixture goal through both paths and asserts the produced HTML matches — far cheaper than catching the drift visually later.

**3. Briefing-dismissal rewire — correct, one theoretical edge.**  `dismissMissionBriefing` now fires from `drainOnClose` gated on `notification.mission_active_gestalt` (set only for briefings, not complete) — correct mapping.  Edge: if a briefing is cancelled during the stub/delay window, `drainOnClose` still runs and marks it seen even though it never displayed.  Briefings carry no `delay` (0 ms), so the window is one macrotask — unreachable in normal play, reachable only via the test helper's synchronous queue drain (which pre-marks briefings seen anyway).  Acceptable; a one-line comment on the cancelled-before-shown path would help a future reader.

**4. `Mark sagt:` vs `Mark says:` cross-component i18n inconsistency (inherited, minor).**  `missionView.ts` falls back to `gettext('Mark sagt:')` (matching its deleted template); the tier-2 components use `gettext('Mark says:')`.  Whichever msgid isn't in the catalog returns the literal string + a console warning.  Inherited from #284, rarely triggered (the ruleset usually supplies `says`).  Worth a catalog-key audit across the ported components — not a #291 blocker.

### Verdict

Clean tier-3 PR.  The typed dispatcher is well-built and closes the #284 follow-up.  No blockers.  I'd want #1 made type-safe (cheap) and a sync-guard test for #2 given its planned multi-tier lifetime; #3 and #4 are notes.

---

PR #294, opened 2026-05-17
Compact the persisted delta log: wire codec + game_values key migration

experintellia:
**Stacked on #292** (base = `claude/fix-storage-quota-error-pC2Ga`). #292 fixes
the storage-quota crash via IndexedDB and persists raw verbose deltas; this PR
layers the size optimization on top and explores how much further to push it.

## Why

#292 stops the quota crash but the append-only log (replayed from serial 0 on
every cold start) is still verbose: the same long keys (`kind`/`addr`/`op`/…
and every `game_values` counter name) repeat on every record. Generic stream
compression is not an option — Delta Chat exposes no deflate on the update
channel, and compressing only inside the dev simulator wouldn't help real
clients. The lever is the *keys*.

## Commits (single-topic, stacked)

| Commit | Topic |
|---|---|
| `feat(state): compact persisted delta log wire codec` | Envelope → single-letter keys + numeric op codes. `decodeDelta` reads legacy verbose **and** compact; `encodeDelta` emits compact only. Wired at the single choke points (LocalEngine send / `applyDelta` read); the fake-webxdc test harness decodes the on-wire payload back to verbose. Append-only wire contract + drift guard. |
| `feat(state): compact game_values keys …migration` | `result.game_values` counters → numeric codes. Read-both / write-short; unmapped/dynamic counters pass through verbatim. |
| `test(state): benchmark for the shorter-key exploration` | Committed deterministic size study (no production wiring) measuring the marginal gain of progressively more aggressive key elimination, to decide how far to go. |

## Migration / backward compatibility

The Delta Chat history is immutable and append-only, so a returning player
replays a mix of old verbose and new compact deltas. Both codecs are
**read-both / write-short**: `decodeDelta` expands compact codes, passes legacy
verbose deltas and unmapped/dynamic `game_values` keys through untouched, and is
idempotent; `encodeDelta` only ever emits the short form. `DELTA_FIELD` /
`OP_CODE` / `GAME_VALUES_FIELD` are append-only contracts (drift tests fail if
an op gains no code).

## Measured (1,000-delta realistic session)

Shipped codec: verbose **356,890 B → 276,994 B (−22.4%)** per-record;
`chargePerp` heavy delta **508 → 366 B (−28%)**.

Shorter-key exploration (marginal vs the shipped codec):

| Variant | vs verbose | marginal vs shipped |
|---|--:|--:|
| V1 shipped codec | 22.4% | — |
| V2 + result/`node` key codes | 29.5% | **−9.2%** |
| V3 + `instance_data` field codes | 35.5% | **−16.9%** |
| V4 + drop constant `addr` | 42.2% | **−25.6%** |
| V5 positional tuple | 49.0% | **−34.2%** |

**Recommendation:** V2 (and likely V3) is the sweet spot — same proven
append-only numeric-code pattern as `GAME_VALUES_FIELD`, meaningful savings,
near-zero risk. V4 needs a merged-multi-peer-history decision; V5 is biggest but
most fragile against the immutable mixed-history contract — not recommended
without versioning. msgpack rejected (no cross-record dictionary; binary breaks
the JSON wire).

## Test plan

- [x] `tsc --noEmit` clean
- [x] Unit: **730/730** (round-trip, mixed-history, open-index, drift, codec↔applyDelta with pinned clock, + the bench)
- [x] e2e regression cluster green with the codec wired in
- [x] Stacked tree verified: identical to #292 + the codec layer + the benchmark, by hash equality

https://claude.ai/code/session_016AEP6a1aBKiiBAMNnxvfmH

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@types/node": "^22.0.0",
     "@vitest/coverage-v8": "^3.0.0",
     "@webxdc/types": "^2.1.2",
-    "@webxdc/vite-plugins": "^1.6.0",
+    "@webxdc/vite-plugins": "github:experintellia/vite-plugins#claude/indexeddb-webxdc-simulator-Oy3ii",
     "lefthook": "^2.1.6",
     "typescript": "^5.3.0",
     "vite": "^8.0.10",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -28,8 +28,8 @@ importers:
         specifier: ^2.1.2
         version: 2.1.2
       '@webxdc/vite-plugins':
-        specifier: ^1.6.0
-        version: 1.6.0(terser@5.46.2)(vite@8.0.10(@types/node@22.19.17)(esbuild@0.27.7)(terser@5.46.2)(yaml@2.8.3))
+        specifier: github:experintellia/vite-plugins#claude/indexeddb-webxdc-simulator-Oy3ii
+        version: https://codeload.github.com/experintellia/vite-plugins/tar.gz/ae3978e30fa35569083bfc16dac3746209eeeb96(terser@5.46.2)(vite@8.0.10(@types/node@22.19.17)(esbuild@0.27.7)(terser@5.46.2)(yaml@2.8.3))
       lefthook:
         specifier: ^2.1.6
         version: 2.1.6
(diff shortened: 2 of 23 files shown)


experintellia:
Closing. #292 (the actual storage-quota fix — IndexedDB persistence + canonical async webxdc API) is merged, so this stacked size-optimization is no longer needed as an active PR.

The codec design, the read-both/write-short migration, the measured compression numbers, and the shorter-key exploration (V2/V3 recommendation, V4/V5 caveats, msgpack rejection) are captured as an idea in #295. The full implementation + the deterministic size benchmark remain on this branch (`claude/explore-shorter-codec-keys`) and can be revived per #295 if log size becomes a concern again.

Not merging.

---

PR #296, opened 2026-05-17
feat(popups): port Contact perp dialog to Preact + perp event-seam keystone (phase 2 tier 5a)

experintellia:
## Summary

Tier 5a (first half of the simple perps): the **Contact** perp dialog, plus the reusable **perp event-seam keystone** that unblocks all of tiers 5–7.

3 commits, each independently validated:

| Commit | What |
|---|---|
| `9a539a0` | `PreactDialogHandle` becomes a real event emitter — `.on(ev, fn)` + general `.trigger(ev, args)` + `close(cb?)`. Lets `GameNode.initPopupEvents` bind `button_click.X` handlers to a Preact-managed popup **unchanged**. `popup_close`/`no_cash`/`no_AP`/`error` keep their special behavior; the registry just *also* dispatches to `.on()` listeners. |
| `c468360` | `GamePerp.openPreactPopup` — perp-owned mount: into the ViewMap `.PopupContainer`, parks the handle on the *perp's* `renderPopup` (so `Charge()`/`collect()` read it), calls `initPopupEvents()` to bind the seam. |
| `b142fb9` | Contact dialog: `contactView.ts` VM + `ContactPopup.tsx` + `ContactPerp.openPopup` reroute; delete `views/popup_contact.html`. |

## Event seam (the architecturally significant bit)

The legacy flow was: jQuery `.Button` delegated handler → `popup.trigger('button_click.ChargeButton')` → `GameNode.initPopupEvents` → `gnode.Charge()` → writes back via `gnode.renderPopup.trigger('no_cash'|'popup_close')`.

Preserved exactly. The Preact button's `onClick` fires `popup.trigger('button_click.ChargeButton')` and parks the clicked element as `popup.lastButton`. No engine logic (`Charge`/`collect`) changed — they read the same handle, now an emitter. The dialog manager injects the live `popup` handle alongside `onClose` (presentational components ignore it).

**Proven end-to-end:** phase-1 Section L (`ChargeButton starts the charge cycle`, `CollectButton queues a profileset after charge ready`) passes against the Preact Contact popup — that exercises click → emitter → `initPopupEvents` → `Charge()`/`collect()` → `popup_close`/`no_cash` back through the handle.

## Token grid

Full port (your call over keeping the partial): `contactView.ts` ports `profileset.html` + `token.html` + `subpop_token.html` — token tiles (stacked `renderSpriteHtml` + `renderAmountHtml`, supertoken/locked/new, `crlf2html` label), **12/page pagination** (`useState`), token→detail-subpop reveal (Preact state, not the legacy jQuery reveal), summary (profiles + risk Up/Down). Legacy partials stay on disk (shared with `popup_project`/`popup_client`); logic duplicated in TS during the transition, same pattern as `missionView.ts`.

## Validation

- [x] `pnpm test` — 705 unit
- [x] `pnpm exec playwright test tests/e2e/dialog-lifecycle.spec.ts` — 43 passed, incl. Section D (Contact open/close), **Section L (charge/collect)**, Section M/O (subpop close), all other perps unaffected
- [ ] Reviewer smoke: click a Contact perp → token grid renders (sprites/labels/amounts), paginate if >12 tokens, click a token → detail subpop opens + closes, Charge (cost decorator, disabled when not idle) and Collect cycle work, popup closes on charge success

## Note — Client/Pusher/Proxy

Originally scoped 5a = Contact+Client. Narrowed to **Contact-only** here so the risky emitter keystone lands proven on the smallest surface. Client (5b) reuses this infra directly (shares the Charge/Collect lifecycle, `profileset_client.html`); Pusher/Proxy (5c, harder — live `fetchProvided` refetch) follow.

## Progress (issue #80)

| Tier | Status |
|---|---|
| 1 status / 2 notifications / 3 missions / 4 About | ✅ #280 #284 #290 #291 #293 |
| 5a Contact + perp seam keystone | **this PR** |
| 5b Client · 5c Pusher/Proxy · 6–8 | next |

11 of 22 dialogs ported.

---


experintellia:
## Review — tier 5a (Contact perp + perp event-seam keystone)

Verified `ContactPopup` / `contactView.ts` against the deleted `popup_contact.html` and the **still-live** shared partials (`profileset.html`, `token.html`, `subpop_token.html` — kept on disk, shared with the not-yet-ported popup_project / popup_client).

### Event seam keystone — sound (positive)

`PreactDialogHandle` gaining `.on()` + emitter `.trigger(ev,args)` + `close(cb?)` faithfully reproduces the legacy `popup` seam. `GamePerp.openPreactPopup` sets `this.renderPopup = handle` **before** `initPopupEvents()` (correct ordering — the seam binds off `renderPopup`), and `onAfterClose` deletes `renderPopup` only if it still `=== handle` (no clobber of a newer popup). The `e.stopPropagation()` in the subpop close handler is load-bearing — it prevents the manager's container `handleInteraction` from double-handling the click — and it's correct. This is a clean keystone; it unblocks 5b–7 as intended.

### Findings

**1. Concrete logic bug — token subpop `collect_amount` source mismatch.** Legacy `subpop_token.html` does `var collect_amount = D.collect_amount || token.collect_amount;`, and `profileset.html` threads the **contact-level** `collect_amount: data.collect_amount` into every token subpop. `contactView.ts:buildToken(token)` has no access to the contact-level value and uses **only `token.collect_amount`**. This changes both:
- the **gate**: legacy `if (collect_amount)` (contact's — almost always truthy for a chargeable contact → *findings* branch) vs Preact `if (collectAmount && findingsText)` where `collectAmount = token.collect_amount` (often undefined per-token → falls through to the *knowledge* branch or nothing);
- the **number**: legacy `toKSNum(contactCollectAmount * token.amount/100)` vs Preact `toKSNum(token.collect_amount * token.amount/100)`.

Net: the Contact token-detail subpop likely shows a different subtitle (knowledge vs findings text) and/or a different value than master. Fix: thread `data.collect_amount` into `buildToken` and mirror `D.collect_amount || token.collect_amount`.

**2. Re-entrancy regression — `fireAction` drops the legacy `:not(.active)` guard.** The legacy `.Button` handler matched `'.Button:not(.disabled, .active)'` and added `.active` on click, structurally blocking re-fire until the popup closed/re-rendered. `ContactPopup.fireAction` checks only `.disabled`, never sets `.active`, and the component never re-renders (snapshot VM; `popup.render()` is a no-op). If Charge/Collect is enabled at open, a rapid double-tap fires `button_click.ChargeButton` twice → `gnode.Charge()` invoked twice before any close. Whether that double-charges depends on `Charge()`/`collect()` re-entry guards — but the legacy structural guard is gone. Recommend an in-flight latch (e.g. a `useRef` "firing" flag, or set+check `.active` on `currentTarget`) or confirm engine idempotency under rapid re-entry.

**3. Extra `<span>` wrapper around the token sprite stack (DOM deviation).** `token.html` injects `RenderSprite(bg)+RenderSprite(perp)+RenderAmount(...)` as **direct children** of `.PopupTokenPerp`. `TokenTile` nests them in `<span dangerouslySetInnerHTML={{__html: token.spriteHtml}} />` (forced because `.PopupTokenPerp` also holds the `.new` ribbon as a real element — Preact disallows children alongside `dangerouslySetInnerHTML`). Any `.PopupTokenPerp > .RenderSprite` / `:nth-child` CSS now sees one `<span>` instead of the sprite layers. Since `token.html` is still live for popup_project/popup_client, Contact tiles can render differently from Client/Project tiles under shared CSS. Consider moving the `.new` ribbon into the sprite HTML string so the stack can be the direct innerHTML of `.PopupTokenPerp`.

**4. Pagination & subpop DOM-model change vs still-live `profileset.html`.** Legacy renders all pages as sibling `.PopupPage` divs slid via the jQuery `.Pagination` handler, and all token subpops pre-rendered in `.SubpopContainer` toggled by `.open`. Preact renders only the current page's tokens in one `.PopupPage` and only the active subpop. Functionally cleaner, but `profileset.html` is still live for popup_project/popup_client → Contact (swap model) and Client/Project (slide model) share `css/Render.css` for `.PopupPageWrap`/`.PopupPage`/`.Subpop`, so pagination/subpop UX diverges during 5a–5c. Also: phase-1 Section K (pagination) exercises the legacy slide model on a template popup — it does **not** cover the Preact Contact pagination, and no new pagination test was added. Coverage gap for the new model; recommend a Contact-pagination e2e test given the divergence will live ~3 tiers.

**5. Touch-path redundancy (minor, latent).** On `touchend`, the manager's legacy `.SubpopClose`/`OKButton` `handleInteraction` still fires for Preact subpops (the Preact `onClick` `stopPropagation` only guards the later synthetic `click`), running redundant jQuery class-manip on Preact-owned DOM before Preact's re-render unmounts it. Benign today (the re-render is authoritative) but a latent two-systems hazard if that jQuery branch ever gains side effects.

**6. `evStub` / no gestalt-data forwarding (forward-looking).** `fireAction` forwards no `[gestalt, data]` and handlers receive a no-op `{stopPropagation, preventDefault}` stub. Fine for Charge/Collect (no `data-button-data`, handlers only stop/prevent). But tier 5b+ buttons that carry `data-button-data` (buy/upgrade) will need `fireAction` extended to read the attributes and forward them, and a richer event object if any handler reads `e.target`.

### Verdict

The event-seam keystone is well-built and is the right shape to carry 5b–7. **Finding #1 is a concrete behavioral bug** (wrong subpop subtitle/value) and should be fixed before merge; **#2 is a real re-entrancy regression** worth a guard or an explicit idempotency confirmation. #3–#6 are equivalence/coverage risks tied to the shared-partial transition — worth a tracked follow-up, especially a Contact-pagination test.

---

experintellia:
## Re-review (head `fcb6695`, 13 commits since the prior review)

Tracking my earlier findings against the new commits.

### Resolved — credit

- **Subpop DOM-model (prior #4, subpop half): fixed.** Commit `132ba65` switches from conditionally unmounting the active `<TokenSubpop>` to rendering **all** token subpops upfront in `.SubpopContainer` and toggling only `.open` — now matches legacy `profileset.html` exactly. Bonus: this fixed a real white-flash bug (the empty absolute container lingered ~0.4s over the grid during the CSS transition). Good catch driven by the model mismatch I flagged.
- **FX feedback for Preact perp dialogs: genuinely good catch (yours).** The legacy `FXNoCash/FXNoAP/FXError` spawned a `RenderSprite` on the board layer, which is occluded by the popup overlay — so a Preact perp dialog had *no error feedback at all*. The DOM-ported `.FXBling` + the all-`.Button` state reset + the `resetMarked()`-on-`animationend` (fixing the "button disabled forever" bug in `0698981`) are the right fixes. The `applyFxFeedback` extraction and the `.FXBling`-orphan-on-early-close reap in `close()` are clean.
- Keystone (prior #5) remains sound; forward-looking #7 (gestalt/data forwarding, shared perp-button hook) is explicitly acknowledged and deferred to 5b in the `/simplify` commit — fine.

### Still open

**1. `collect_amount` source mismatch (prior #1) — NOT fixed, still the top concrete bug.** `contactView.ts:buildToken(token)` still uses only `token.collect_amount`. Legacy `subpop_token.html` uses `D.collect_amount || token.collect_amount` where `profileset.html` threads the **contact-level** `collect_amount: data.collect_amount`. `data.collect_amount` is right there in `buildContactPopupVM` (it's already used for `summaryProfiles`) but isn't passed into `buildToken`. Net: the token-detail subpop subtitle takes a different branch (findings vs knowledge vs empty) and, where findings shows, a different multiplier base, vs master. The `/simplify` commit listed deliberately-skipped non-issues — this wasn't among them, so I think it was just missed rather than dispositioned. Two-line fix: thread `data.collect_amount` into `buildToken` and mirror `D.collect_amount || token.collect_amount`. Please fix or explicitly say why it's not a regression.

**2. Token sprite `<span>` wrapper (prior #3) — NOT fixed.** `TokenTile` still nests the sprite stack in `<span dangerouslySetInnerHTML>` inside `.PopupTokenPerp`, where the still-live shared `token.html` injects the sprites as direct children. `.PopupTokenPerp > *` / `:nth-child` CSS will see one `<span>` instead of the layers — Contact tiles can diverge from the still-legacy Client/Project tiles under shared CSS. Moving the `.new` ribbon into the sprite HTML string lets the stack be the direct innerHTML of `.PopupTokenPerp`.

**3. Pagination model (prior #4, pagination half) — still divergent, still untested.** Still single-page-swap (`pageTokens.map` in one `.PopupPage`) vs legacy all-pages-slide in the still-live `profileset.html`. No Contact-pagination e2e added (commit count stays "43 dialog e2e"). Given this divergence lives until 5c, a Contact >12-token pagination test is worth adding now.

### Improved but note

**4. Re-entrancy (prior #2) — materially mitigated, not structurally closed.** `applyFxFeedback` now clears state from every `.Button` and disables the marked one on `no_cash`/`no_AP`/`error`, so the post-failure re-click window is covered, and the synchronous-close path covers success. The explicit legacy `:not(.active)` guard is still absent, so the only remaining exposure is a rapid double-tap *before* a (hypothetically async) `Charge()` resolves. Fine for the synchronous local engine today; worth a one-line comment that engine-side idempotency is the sole guard for that window.

### New finding

**5. `applyFxFeedback` strips `.disabled` from *all* buttons and Preact never restores it.** `for (const b of container.querySelectorAll('.Button')) b.classList.remove('active','disabled','no_cash','no_AP','ERROR')` removes the class from every `.Button`, including one the VM intentionally rendered `disabled` (`vm.chargeDisabled` → `class="Button disabled"`). Since the component never re-renders (snapshot VM, `popup.render()` is a no-op), a VM-disabled button that gets cleared here stays wrongly enabled until the popup closes. Low exposure for Contact (one action button rendered at a time, and it's the one clicked) — but a real hazard for tier-5b+ perp popups with multiple conditionally-disabled buttons. Recommend `applyFxFeedback` only reset the FX classes (`no_cash`/`no_AP`/`ERROR`/`active`) and leave `disabled` to the VM, or have it restore VM-disabled state.

### Verdict

Good progress — the subpop model fix and the FX-bling port are solid. **Finding #1 is still a live behavioral regression and the one I'd block on**; the new #5 is worth fixing before this becomes the template for 5b–7 (multi-button perp popups). #2/#3 are equivalence risks tied to the shared-partial transition; #4 is a note.

---

PR #297, opened 2026-05-17
Simplify AboutPopup: remove debug tab, extract links to i18n

experintellia:
## Summary
Simplifies the AboutPopup component by removing the debug tab functionality and extracting hardcoded links and text into i18n strings. The dialog is now single-purpose and stateless, reflecting that webxdc has no reset functionality.

## Key Changes
- **Removed debug tab**: Eliminated the conditional tab strip and Debug tab that was gated on `userdebug`. The legacy reset functionality doesn't apply to webxdc (reset = re-share the .xdc file).
- **Extracted text to i18n**: Converted all hardcoded English prose into translatable i18n strings with `%s` placeholders for external links.
- **Created Link component**: New `Link.tsx` component centralizes external link styling (`class="mln"`) and security attributes (`target="_blank" rel="noreferrer noopener"`).
- **Implemented link splicing**: Added `withLinks()` helper function to splice `<Link>` components into i18n template strings at `%s` markers.
- **Updated props**: Removed `userdebug` prop from `AboutPopupProps` since it's no longer needed.
- **Removed useState**: Eliminated tab state management since there's now only one tab.
- **Updated tests**: Removed e2e test for tab switching behavior and updated comments about the dead ResetButton.
- **Added i18n strings**: Added German and English translations for all About dialog content (fork headline/text, displayname text, original game headline/text, FAQ link).

## Implementation Details
- The `withLinks()` function splits i18n strings on `%s` markers and interleaves them with corresponding `<Link>` components, maintaining proper ordering.
- The dialog remains purely informational with only a Close button, matching the webxdc use case.
- All external links now use the centralized `Link` component for consistency.

https://claude.ai/code/session_015JLEWuccHojZTkMVsJLyUR

### DIFF
--- a/i18n/de_AT.json
+++ b/i18n/de_AT.json
@@ -693,6 +693,30 @@
     null,
     "Über"
   ],
+  "about fork headline": [
+    null,
+    "Über diesen Fork"
+  ],
+  "about fork text": [
+    null,
+    "Dies ist eine webxdc-Portierung des originalen Data-Dealer-Browserspiels, neu verpackt, damit es ohne Server in einer Delta-Chat-Gruppe läuft. Quellcode, Fehlerberichte und Beiträge: %s."
+  ],
+  "about displayname text": [
+    null,
+    "Dein Anzeigename wird über dein Delta-Chat-Profil festgelegt — ändere ihn dort, und das Spiel übernimmt ihn automatisch."
+  ],
+  "about original headline": [
+    null,
+    "Das Originalspiel"
+  ],
+  "about original text": [
+    null,
+    "Data Dealer wurde zwischen 2011 und 2014 in Wien von Cuteacute Media OG entwickelt (Ivan Averintsev, Wolfie Christl, Pascale Osterwalder, Tobi Schäfer, Ralf Traunsteiner). Es war ein satirisches Browserspiel über den Handel mit persönlichen Daten, das als öffentliche Beta auf datadealer.com veröffentlicht und als %s, %s und %s quelloffen gemacht wurde."
+  ],
+  "about original faq": [
+    null,
+    "Hintergrundinfos und die ursprüngliche FAQ: %s."
+  ],
   "agent_popup selector title": [
     null,
     "Kontakte anwerben"
--- a/i18n/en_US.json
+++ b/i18n/en_US.json
@@ -693,6 +693,30 @@
     null,
     "About"
   ],
+  "about fork headline": [
+    null,
+    "About this fork"
+  ],
+  "about fork text": [
+    null,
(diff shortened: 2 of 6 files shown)


experintellia:
## Review — Simplify AboutPopup

Reviewed the `withLinks()` splice, the i18n catalog changes, link security, and the test-contract impact.

**`withLinks()` correctness.** Sound for all three call sites (`%s`-count == link-count, no leading/adjacent `%s`), but it's a reusable helper with latent edge-case behavior and no dev-time guard:

```js
const parts = template.split('%s');
parts.forEach((part, i) => {
  if (part) out.push(part);
  const link = links[i];
  if (link) out.push(link);
});
```

- `if (part)` silently drops empty segments — a *leading* `%s` or adjacent `%s%s` mis-splices.
- More `%s` than links → trailing placeholder slot silently consumed (a `de` string with an extra `%s` drops a link with no visible error).
- More links than `%s` → extra links silently dropped (`links[i]` never reached).

None fire for the current controlled inputs, but recommend a `links.length !== parts.length - 1` dev `console.warn` before this gets reused with translator-supplied strings. Keys are handled — each `<Link>` carries an explicit `key=` at the call site, so the mixed string/element array won't trigger Preact reconciliation warnings (a self-keying helper would be more robust, but it's fine).

**i18n.** Verified all 6 new msgids (`about fork headline/text`, `about displayname text`, `about original headline/text/faq`) are in **both** `i18n/en_US.json` and `i18n/de_AT.json`, and the `%s` counts match between en/de and equal the spliced link counts (fork text = 1, original text = 3, original faq = 1, others = 0). No literal-fallback / mismatched-placeholder risk. Note: only the compiled `.json` was touched, not the `.po` sources — confirm the i18n pipeline treats `.json` as canonical so the `.po` doesn't silently drift.

**Link security — improvement, correctly done.** `Link.tsx` pairs `target="_blank"` with `rel="noreferrer noopener"` and `class="mln"`. No `dangerouslySetInnerHTML` anywhere — the old inline `&mdash;`/`&auml;` HTML entities are replaced with literal Unicode in the i18n strings, so prose now renders as text nodes. Net XSS-surface reduction.

**Test contract — clean.** The removed e2e test was the only assertion referencing `.PopupBody.About .PopupMenuButton[data-tab="debug"]` / `.PopupTab[data-tab=...]`. Section C (settings/About) only does open/close (`.PopupBody.About` visible → `expectOpenAndClose`) — still passes. No other About-scoped tab/ResetButton assertions remain; the Section N dead-code audit comment was updated, not relied on by a test. `GameRoot.ts` correctly drops the `userdebug` prop. No orphaned assertions.

### Verdict

Approve. Behavior is correct for all current inputs and the test contract stays consistent. Only real concern: `withLinks()` silently swallows mismatched `%s`/link counts with no dev warning — harden it before it's reused with translator-supplied strings (`AboutPopup.tsx` `if (part)` + the missing length check).

---

PR #300, opened 2026-05-17
Fix charge-ready emission to use entry data instead of state

experintellia:
## Summary
Refactored `_scheduleChargeReady()` to emit `node_ready` directly from the charging entry data rather than re-deriving it from global state. This prevents a race condition where the chargePerp echo may not have been applied when the timer fires, causing the perp to appear stuck in "charging" state until reload.

## Key Changes
- **Updated function signature**: `_scheduleChargeReady()` now accepts the full `ChargingEntry` object instead of separate `chargeEnd` and `path` parameters
- **Direct event emission**: Replaced the generic materialize+event-loop approach with a direct `node_ready` emission using the entry's own data (`game_id`, `game_type`, `path`, `result`)
- **Removed stale-state check**: Eliminated the `nodes_charging` lookup that would fail if the echo hadn't landed yet
- **Simplified logic**: Removed unnecessary event filtering; now emits exactly one `node_ready` event per charge cycle
- **Updated call sites**: Modified both `loadGame()` and `chargePerp()` to pass the entry object directly

## Implementation Details
The key insight is that under canonical-async webxdc, messenger lag or app backgrounding can cause the chargePerp echo to arrive after the charge timer fires. By using the entry data the caller already holds (rather than re-reading from `nodes_charging`), we guarantee the emit succeeds regardless of echo timing. The `materialize()` call still advances time-based state (AP regen, eventual charge→collect promotion), but the critical `node_ready` event is now decoupled from state consistency, eliminating the stuck-action scenario.

https://claude.ai/code/session_01HP1ABvsFgnYWBeTCqryaox

### DIFF
--- a/scripts/LocalEngine.ts
+++ b/scripts/LocalEngine.ts
@@ -1576,7 +1576,13 @@ export function buyPerp(
   var levelup = newLevel > oldLevel;
   if (levelup) newGv = _applyLevelUp(newGv, newLevel);
 
-  var missionResult = _advanceBuyPerpMissions(state, gestalt);
+  // Project the just-bought node in before the mission cascade so a
+  // cascade-unlocked buy_perp goal targeting this perp auto-completes now
+  // instead of staying stuck until a reload replays the buy delta.
+  var preMissionStateBuy = Object.assign({}, state, {
+    nodes: state.nodes.concat([newNode]),
+  });
+  var missionResult = _advanceBuyPerpMissions(preMissionStateBuy, gestalt);
   newGv = _applyRewardsToGv(newGv, missionResult.rewards);
 
   // profile_set for project*/contact*/city* gestalts:
@@ -1782,7 +1788,7 @@ function _advanceIntegrateProfilesMissions(
     return { missions: null, mission_goals: goals, active_missions: activeMissions };
   }
 
-  return _completeMissionsIfReady(updatedGoals, activeMissions);
+  return _completeMissionsIfReady(updatedGoals, activeMissions, state.nodes);
 }
 
 function _advanceCollectProfilesMissions(
@@ -1814,12 +1820,13 @@ function _advanceCollectProfilesMissions(
     return { missions: null, mission_goals: goals, active_missions: activeMissions };
   }
 
-  return _completeMissionsIfReady(updatedGoals, activeMissions);
+  return _completeMissionsIfReady(updatedGoals, activeMissions, state.nodes);
 }
 
 function _completeMissionsIfReady(
   updatedGoals: MissionGoal[],
-  activeMissions: string[]
+  activeMissions: string[],
+  nodes: GameNode[]
 ): MissionAdvanceResult {
   var ruleset = _getRuleset();
   var completed: string[] = [];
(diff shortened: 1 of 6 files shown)


experintellia:
## Review — charge-ready emission from entry data

Reviewed the fix and traced both `_scheduleChargeReady` call sites (`chargePerp()`, `loadGame()`).

### The race is genuinely closed

Both call sites pass a stable, fully-populated `ChargingEntry` (`path`, `result`, `charge_start`, `charge_end`, `game_id`, `game_type`). `ChargingEntry` is never mutated in place anywhere in the engine (entries are replaced via `_filterByPath`/`concat`), so the timer closure's captured `entry` stays valid. The emit path no longer reads `nodes_charging` — the stale-state probe is gone and the `node_ready` payload is built purely from `entry`. The echo-lag stuck-charge scenario cannot recur.

### `materialize()` decoupling is sound, double-emit is handled

The genuinely risky path is the `loadGame()` replay: a charge that completed during the away window is emitted once by `mat.events` (the `queueMicrotask` loop) **and** by the re-armed timer (`msUntil = max(0, charge_end - now)` = 0 → fires next tick, after the microtask). This is neutralized by the `markReady()` idempotency guard (`if (gperp.renderReady) return;`) present in all four perp classes — the second `node_ready` is a UI no-op. Without that guard this PR would regress; with it, correct. Worth a one-line comment at the re-arm site noting the idempotency guard is what makes the double-emit safe, since that coupling is non-obvious.

### Timer lifecycle is safe

`_clearAllChargeReady()` runs at the top of every `loadGame()`, so a reload before the timer fires can't leak a stale timer. The captured-entry emit can't error if the perp/popup is gone — `node_ready` handlers just call the (now idempotent) `markReady()`.

### One required change: add the regression test

`changed_files` is 5 (engine + the four perp classes) with **no spec touched**. A concurrency fix shipping with zero coverage of the reproduced bug is the one real deficiency. Add a test asserting `_scheduleChargeReady`'s timer emits `node_ready` when `nodes_charging` is empty at fire time (the exact echo-lag scenario), and ideally one pinning the `loadGame()`-replay no-double-fire (timer + microtask, guard makes it a single UI effect).

### Verdict

Fix is correct and the `markReady` idempotency guard properly neutralizes the replay double-emit — approve, but it should not merge without a test reproducing the echo-lag stuck-charge it fixes.

---

experintellia:
Addressed the merge-blocking item in `560ee62`.

- Added `tests/handlers/chargePerp.test.js` → *"emits node_ready from the captured entry even when nodes_charging is empty at fire time"*: arms the timer via `chargePerp`, then drops the charge from live state (simulating the canonical-async echo not yet applied) and asserts exactly one `node_ready` fires with the right path/result. This fails on pre-fix code (the `if (!stillCharging) return;` probe bailed) and passes now.
- Added the one-line comment at the `loadGame` re-arm site noting the queueMicrotask / re-armed-timer overlap is a UI no-op via the `markReady()` idempotency guard.

On the *ideally* second test: at the engine level a charge that completes during the away window is removed from `nodes_charging` by `loadGame`'s `materialize()` *before* the re-arm loop, so it's emitted only by the `queueMicrotask` `mat.events` path and no timer is re-armed for it — the timer/microtask paths are mutually exclusive, so there's no engine-level double-fire to pin; the `markReady` guard remains the belt-and-suspenders for the `_loadReady` overlap.

709 unit tests + typecheck + Biome green.

---

PR #301, opened 2026-05-17
Refactor FX feedback handling and extract toFXTarget utility

experintellia:
## Summary
This PR refactors the FX feedback system in the dialog manager to reduce code duplication and improve maintainability. The main changes involve extracting a shared `toFXTarget()` utility function and consolidating the button targeting logic for visual feedback effects.

## Key Changes

- **Extract `toFXTarget()` utility**: Created a new exported function in `dialogManager.ts` that wraps DOM elements as `FXClassTarget` objects. This replaces the duplicate `asLastButton()` implementation in `ContactPopup.tsx`.

- **Consolidate button targeting logic**: Unified the `lastButton` and fallback MainButton handling in `applyFxFeedback()` by introducing a `target` variable that represents either the clicked button or the main button fallback, eliminating nested conditionals.

- **Convert `FX_RESET_CLASSES` to array**: Changed from a space-separated string to an array of class names for better type safety and clarity, with corresponding updates to how classes are removed (using `.join(' ')` instead of `.split(' ')`).

- **Simplify class manipulation**: Updated class removal logic to work consistently with the new array format and the unified target approach.

- **Update `buildToken()` signature**: Changed the function to accept the full `ContactData` object instead of just `contactCollectAmount`, allowing it to access `contact.collect_amount` directly. This reduces parameter coupling and makes the intent clearer.

- **Refine comments**: Removed outdated or overly detailed comments about legacy behavior and clarified remaining documentation.

## Implementation Details

The `toFXTarget()` function provides a consistent interface for adding/removing space-separated CSS classes on DOM elements, handling empty class names gracefully. This allows both the popup's clicked buttons and the dialog manager's MainButton fallback to use the same FX feedback mechanism without code duplication.


experintellia:
## Review — tier 5b (Client perp port + shared perp extraction)

Note: the PR title ("Refactor FX feedback / extract toFXTarget") undersells this — it's the full tier-5b Client port (new `ClientPopup.tsx`, `clientView.ts`, shared `tokenView.ts` + `perpShared.tsx`, `popup_client.html` deleted) plus the 5a `/simplify`. Reviewed against the deleted `popup_client.html` and the legacy `profileset_client.html` / `token_consumed.html`.

### Carried-forward findings — correctly resolved (credit)

- **Prior #296 #1 (collect_amount source)** is correctly generalized: `tokenView.buildSubpop` uses `ctx.collect_amount || token.collect_amount` with the legacy `if (collect_amount) … else if (database_amount)` gate, and both Contact and Client thread the perp-level `data.collect_amount` as `ctx`. Matches `subpop_token.html` via `profileset[_client].html`. ✓
- **Prior #296 #5 (applyFxFeedback stripping `.disabled` from all buttons)** survives the refactor — the cross-button reset is now only `active/no_cash/no_AP/ERROR`, `disabled` is left to the VM. ✓
- The `perpShared.tsx` / `tokenView.ts` extraction is the "extract when 5b adds the 2nd consumer" deferred in the 5a `/simplify` — clean follow-through. `buildConsumedToken` vs `token_consumed.html` verified faithful (pivot math, `'consumed'` frame on bg + amount, no ribbon, perp_sprite at normal frame, `database_amount && !locked` gate). `buildToken` provided-set mapping matches `profileset_client.html` (the legacy `matching:true` param is dead — `token.html` never reads it). Good test additions (Section K Client dual-grid + independent pagination; Section L Client Charge/Collect + `giveAP`).

### Concrete finding

**1. `.PopupButtons` is nested one level too deep vs the deleted `popup_client.html` (DOM-structure regression).** In the legacy template the `.PopupTab` closes **before** the button block — `.PopupButtons` is a direct child of `.PopupContent`, a **sibling** of `.PopupTab`:

```
.PopupContent
  .PopupTab            (profileset_client.html only)
  .PopupButtons        (sibling of .PopupTab)
```

`ClientPopup.tsx` renders `.PopupButtons` (and `.PopupSummary`) **inside** `.PopupTab`. (Note: `.PopupSummary` *is* correct inside `.PopupTab` — `profileset_client.html` emits it there. Only `.PopupButtons` deviates.) This is almost certainly a copy of the Contact structure, where the legacy `popup_contact.html` *did* put buttons inside `.PopupTab` — but `popup_client.html` does not. Net: the Client popup's Charge/Collect bar now sits one nesting level deeper than master, so any `.PopupContent > .PopupButtons` vs `.PopupTab > .PopupButtons` layout/flex rule renders the Client button bar differently from master. Move `.PopupButtons` out of `.PopupTab` (sibling, after it) to match the deleted template, or confirm via the CSS that the two positions are visually identical.

### Verify / cleanup

**2. `profileset_client.html` and `token_consumed.html` look orphaned by this PR.** Their only renderer was `popup_client.html`, which this PR deletes. Unlike `profileset.html` / `token.html` / `subpop_token.html` (genuinely shared — still rendered by the not-yet-ported `popup_project`/mission card), the `_client` and `_consumed` partials are client-only. If nothing else renders them they're now dead weight and should be deleted in this PR (matching how `popup_contact.html` was deleted in 5a). Please confirm whether any live template still references them; if not, delete.

### Minor / faithful-to-legacy notes

**3. Summary income text gains an extra `<span>` wrapper.** `profileset_client.html` prints the non-penalty income as a bare text node in `.PopupSummaryItem`; `ClientPopup` wraps it in `<span dangerouslySetInnerHTML>` (forced — `.PopupSummaryItem` also has the real `.RenderSprite.Tobi` child, so the dual-grid trick used for the token tile can't apply). `profileset_client.html` appears dead after this PR so there's no shared-CSS divergence, but it's a deviation from what master rendered; flagging for visual parity. Low severity.

**4. Provided/consumed gestalt collision opens both subpops.** `key={`${i}-${t.gestalt}`}` correctly avoids the Preact dup-key, but `isOpen={openToken === t.gestalt}` is gestalt-keyed — if a provided and a consumed token share a gestalt, clicking one renders **both** `.Subpop.open`. This faithfully reproduces the legacy behavior (the jQuery reveal did `.find('.Subpop[data-subpop-id=tokenG]')`, which also matched both), so it's not a regression — just noting the preserved latent bug in case 5c wants to fix it.

### Verdict

The shared extraction and the carried-forward fixes are solid. **Finding #1 (`.PopupButtons` nesting) is a concrete DOM regression vs the deleted `popup_client.html`** and is the one to fix before merge; #2 (orphaned templates) is a cleanup/verify; #3/#4 are minor/faithful notes.

---

PR #302, opened 2026-05-17
feat(popups): port Pusher + Proxy perp dialogs to Preact (phase 2 tier 5c)

experintellia:
## Summary

Tier 5c — the **Pusher** and **Proxy** perp buy dialogs. Unlike Contact/Client (Charge/collect), these render a paged "provided perps" buy grid and **live-refetch** (`vclick → fetchProvided → compileProvided → updatePopup`). Both popups share one shell.

| File | What |
|---|---|
| `scripts/game/providedView.ts` | Shared VM porting the partials the Pusher/Proxy (and later Agent/City/Project) buy popups all render: `client.html`/`perp.html` tiles, `subpop_perp_provided.html` (+ `values.html`/`values_details.html`), `noitems.html`. Faithful: pivot math (48 client / 49 perp), supertoken `perp_background2`, `CityPerpSpecial`, locked `Requires` (Pusher: required providers; Proxy: DB-token-filtered + level fallback), the ~12-branch values/bonus blocks. Templates stay on disk (still rendered by not-yet-ported popups) — same transition pattern as `tokenView.ts`. |
| `perpShared.tsx` | `PerpProvidedTile` / `PerpProvidedSubpop` / `NoItems`; `PageArrows` gains `standalone`; `fireAction` forwards `[gestalt]` so `PerpBuyButton` routes through the legacy seam (the deferred 5a review #6). |
| `ProvidedPerpPopup.tsx` + `pusherView.ts` / `proxyView.ts` | One shell; the thin views resolve subtitle / selector title / tile kind / button-disabled / empty-state copy into one `ProvidedPopupVM`. |
| `PusherPerp.ts` / `ProxyPerp.ts` | `openPopup()` reroute; `updatePopup()` = **Path-A re-mount** (openDialog closes + remounts with the fresh VM on refetch, matching the legacy rebuild). Deleted `popup_pusher.html` / `popup_proxy.html` (Pusher/Proxy-only top-level templates). |

## Live-refetch decision

Path A (re-mount) over a hooks/subscription model — keeps the snapshot-VM architecture consistent with 5a/5b and matches the legacy rebuild; the only cost is a brief re-mount on the rare "popup open while the server fetch completes" case.

## Validation

- [x] `pnpm typecheck` (both tsconfigs) / `pnpm lint` / `pnpm build` — clean
- [x] `pnpm test` — 708 unit
- [x] `pnpm exec playwright test tests/e2e/dialog-lifecycle.spec.ts` — 48 passed, incl. **Section P** (provided grid, PerpBuyButton gestalt seam, Proxy slot subtitle + NoItems, Path-A re-mount) and Section D Pusher/Proxy open-close, all other perps unaffected
- [ ] Reviewer smoke: open a Pusher and a Proxy perp → buy grid renders (sprites/labels/price/values), tile → buy subpop with Buy button, 5/page pagination, Proxy "Daughter companies: used/max" subtitle, loading/no-items empty state, refetch refreshes the grid; regression-smoke a Contact/Client perp (shared `perpShared`/`PageArrows`/`fireAction` changed)

## Progress (issue #80)

| Tier | Status |
|---|---|
| 1 status / 2 notifications / 3 missions / 4 About | ✅ |
| 5a Contact + perp seam keystone | ✅ #296 |
| 5b Client + shared tokenView/perpShared | ✅ #301 |
| 5c Pusher/Proxy + shared providedView | **this PR** |
| 6 complex perps · 7 catalogs · 8 cleanup | next |

---


experintellia:
## Review — tier 5c (Pusher + Proxy provided-perp popups)

Diffed `ProvidedPerpPopup` / `providedView.ts` against the deleted `popup_pusher.html` / `popup_proxy.html` and the still-live `client.html` / `perp.html` / `noitems.html`.

### Concrete finding

**1. Proxy `slot_background` fallback mis-positions `.RenderPerp` (visual regression).** `perp.html` computes the `.RenderPerp` offset/size **only** inside `if (perp_background && perp_background.frameMap.normal)`; its `else if (data.slot_background)` branch reassigns `background` for the sprite render **but leaves offsets at the initialized `0,0,100,100`**. `providedView.buildPerpTile` does:

```ts
let bg = spriteOf(data.perp_background);
if (bg && isSuper) bg = spriteOf(data.perp_background2);
if (!bg && data.slot_background) bg = spriteOf(data.slot_background);
... perpStyle(normalFrame(bg), 49) ...
```

So when `perp_background` is absent and `slot_background` is present (the Proxy empty-slot tile), Preact runs the pivot math on `slot_background.frameMap.normal` (→ `top:(49-pivoty); left:(49-pivotx); width/height` from the slot frame), whereas master renders that tile at `top:0; left:0; width:100px; height:100px`. Slot-background sprite configs generally carry a `frameMap.normal`, so this likely shifts/sizes **every Proxy empty-slot tile** differently from master. Fix: mirror the legacy gating — only compute offsets when `data.perp_background?.frameMap?.normal` exists; for the `slot_background` fallback keep `0,0,100,100` (use `slot_background` for the sprite only).

### Verify (not diffed — highest remaining risk)

**2. `buildValuesHtml` / `buildValuesDetailsHtml` (~12 branches each) vs `values.html` / `values_details.html`.** I couldn't diff these against their templates (not fetched). One asymmetry in the diff itself is worth confirming: the **karma** row differs between the two functions — `buildValuesHtml` emits `<div class="Risk Up">…{karma}…karma_image</div>` (always `Up`, label `karma_image`) while `buildValuesDetailsHtml` emits `<div class="Bonus Risk ${karma>0?'Up':'Down'}">…{karma}…Risk</div>` (signed, label `Risk`). That may be faithful (the two legacy partials genuinely differ) or a port slip. Please confirm both branch lists — ordering, the `< 0` vs `> 0` sign tests, and the `%s`-placeholder labels — against `values.html` / `values_details.html`, since this is the most drift-prone surface and the e2e tests don't pin the value-row contents. `subpop_perp_provided.html` structure is likewise unverified.

### Minor

**3. `required_level` absent flips the locked branch.** `perp.html` gates the providers-vs-level branch on `data.required_level <= groot.xp_level.number`; with `required_level` undefined that's `undefined <= N` → `false` → the "Level" fallback. `buildPerpTile` uses `reqLevel = data.required_level ?? 0`, so `0 <= N` → `true` → the providers branch instead. Only matters for a locked perp that has `requiredTokens` but no `required_level` (rare), but it's a boundary divergence introduced by the `?? 0` default; legacy also renders "Level undefined" there, so neither is pretty — worth a deliberate decision rather than an accidental flip.

### Verified faithful / safe (no action)

- **`.PopupButtons` placement is correct here** — `popup_pusher.html` / `popup_proxy.html` put `.PopupButtons` *inside* `.PopupTab`, and `ProvidedPerpPopup` matches. (Contrast the #301 Client finding where the legacy template had it outside — that distinction was handled correctly this time.)
- **`display:contents` wrapper around the values/requires HTML is safe.** `.PowerupLabelData`/`.PerpLabelData` have no child-combinator (`>`) or `:nth/:first/:last-child` rules in `css/Render.css`, so wrapping the rows in a `display:contents` div is layout-transparent and selector-safe. (One forward-looking note: if future CSS adds a `.PowerupLabelData > .Profiles`-style rule it would silently miss the Preact path only — but nothing today depends on it.)
- **NoItems text is exact.** `noitems.html` defaults are `'Currently no items available'` / `'Loading items'`; `popup_pusher.html` passes none → those defaults, and `pusherView` hardcodes exactly those. `popup_proxy.html`'s explicit strings match `proxyView` verbatim. ✓
- **`client.html` (Pusher) tile port is faithful** — outer classes, `data-subpop-id`=index, `data-gestalt`, 48-box pivot, `PowerupBackground`/`PowerupSprite`, `crlf2html` label, `.Price > .Buy.Cash`, locked `Requires`/`RequiresProviders` with `', '` separator all match. `perp.html` locked `RequiresProviders` separator (`',<br />'`), the `!DBTokens.hasOwnProperty` filter, and the supertoken `perp_background2` selection all match (and `buildPerpTile` is more defensive on missing `type_data`/`frameMap` than the throwing legacy — fine).
- **Path-A re-mount is implemented cleanly.** `updatePopup → openPopup → openDialog` closes the old handle (its `onAfterClose` clears `renderPopup`) then mounts fresh; because each `openDialog` owns its own `listeners` Map and the old handle is discarded, `initPopupEvents()` re-binding on every refetch does **not** leak/duplicate `button_click` handlers, and the teardown→mount is synchronous so no flicker. The only cost is `page`/`openKey` resetting on a mid-interaction refetch — confirm the legacy template rebuild also reset pagination/open-subpop (it almost certainly did, since it re-rendered the whole template); if so this is faithful, not a regression.
- **`.PopupBody.ProvidedPerp`** is a net-new modifier class (additive, scopes the new subpop/`.PopupContent` transparency CSS) — phase-1 `.PopupBody` selectors still match; fine. The legacy dev-only `<div class="Debug">` dump is dropped — consistent with the AboutPopup debug removal; flagging only so it's a conscious choice.

### Verdict

Finding #1 (Proxy `slot_background` positioning) is a concrete visual regression vs `perp.html` and should be fixed before merge. #2 (values/values_details branch fidelity) is the key un-diffed surface — please confirm against the two templates, especially the karma-row asymmetry. #3 is a minor boundary note. Everything else verified faithful; the shared `providedView`/`perpShared` extraction and the Path-A re-mount are sound, and the good test additions (Section P incl. both Proxy locked branches + pagination + the gestalt seam) are solid.

---

experintellia:
## Re-review (head `451c6ea`, 7 commits since the prior review)

Tracked my three findings against the new commits + independently re-verified the one substantive item.

### Finding #2 (values / values_details fidelity) — CLOSED, independently confirmed faithful

You dispositioned this "faithful per sub-agent, no code change"; I diffed `values.html` and `values_details.html` line-by-line against `buildValuesHtml` / `buildValuesDetailsHtml` to settle it rather than defer. **Both 12-branch blocks are exact ports** — branch order, element classes, the `< 0` vs `> 0` sign tests, the `%s` labels, and the guards all match:

- `values.html`: tokens → consumed → collect_amount_modifier → collect_amount → charge_cost_modifier → charge_cost(`button_text`) → income_base → collect_risk → collect_risk_modifier → karma(`Risk Up`/`karma_image`) → max_slots → provided_perps. ✓ exact.
- `values_details.html`: consumed → tokens → collect_amount_modifier → collect_amount → **max_slots (before charge here)** → charge_cost_modifier → charge_cost(`per deal`) → income_base → collect_risk → collect_risk_modifier → karma(`karma>0?Up:Down`/`Risk`) → provided_perps(`&& provided_perps_text`), wrapped in `.BonusWrap > .BonusValues`. ✓ exact.

The **karma-row asymmetry I flagged is genuinely faithful** — `values.html` hard-codes `Risk Up` + `karma_image`; `values_details.html` uses signed `Up/Down` + `Risk`; the two legacy partials really do differ and the port reproduces each correctly. Concern withdrawn. (The dead `xp_level`/`checklevel`/`game` locals in `values.html` are correctly dropped — clean port.)

### Finding #1 (Proxy `slot_background` offset) — disposition correct for Proxy, but the fix belongs in this PR

Your call ("Proxy provides only ProjectPerps with no background fields → the fallback edge is unreachable, no code change") is **correct for Proxy's runtime data**: with neither `perp_background` nor `slot_background` set, `bg` stays undefined → `perpStyle(undefined, 49)` → `0,0,100,100`, matching `perp.html`'s untaken-else path.

But `providedView.buildPerpTile` is, by your own file header and commit message, the **shared module for "Pusher/Proxy (and later Agent/City/Project)"**. `slot_background` is precisely the empty-slot field that `perp.html`'s `else if (data.slot_background)` branch exists for — and that City/Project (tier 7) and Agent (tier 6) feed. So this isn't an ordinary latent bug; it's a *known* ~2-line faithfulness gap being shipped into the shared infra that later tiers will route slot-bearing perps through, at which point `buildPerpTile` will apply pivot math where `perp.html` deliberately keeps `0,0,100,100`. Recommend fixing it now — only pass `perpStyle` a frame when `perp_background` (or `perp_background2` if supertoken) has `frameMap.normal`; the `slot_background` fallback should affect the rendered sprite only, never the offset, exactly as legacy. If you'd rather defer, a code comment in `buildPerpTile` marking the known divergence + a tier-6/7 tracking ref is the minimum, so it's not silently rediscovered.

### Finding #3 (`required_level ?? 0` flips the locked branch) — still open (minor)

Unchanged; the commit's "#3" referred to test-coverage gaps (numbering mismatch with my comment). `reqLevel = data.required_level ?? 0` still means a locked perp with `requiredTokens` but no `required_level` takes the providers branch in Preact where `perp.html` (`undefined <= xpLevel` → false) takes the Level fallback. Low-probability; just flagging it remains an undispositioned conscious-decision item, not that it must change.

### Subpop white-flash CSS iteration — well-handled self-found polish

The 5 commits chasing the Pusher/Proxy subpop close flash (`.SubpopContainer` transparent → insufficient → also `.PopupBody.ProvidedPerp .PopupContent` transparent + `.Subpop` opaque only while `.open`) are sound: exploiting `background` not being in the `.Subpop` transition list so it vanishes on the first close frame while transform/opacity still animate is a clean technique, the differential-probe verification is convincing, and scoping to `.PopupBody.ProvidedPerp` correctly leaves Contact/Client and other Preact dialogs untouched. The `8f07957` doc-comment correctly records that still-legacy ProjectPerp/Agent popups keep the legacy flash until ported — a deliberately-scoped, documented tradeoff (avoiding legacy-CSS blast radius), which is the right call. The final comment-trim is consistent with the repo's "don't narrate the investigation" convention. No concerns.

### Verdict

No blockers. **#2 fully closed** (independently confirmed faithful, including the karma asymmetry). **#1** reduced to: the disposition is right for Proxy, but since this PR *is* the defining PR for the shared `buildPerpTile`, the ~2-line faithful gate (or at least a tracked code comment) should land here rather than ship a known trap for tier 6/7. **#3** minor/your call. The CSS polish and the added pagination/locked-tile coverage are solid.

---

PR #311, opened 2026-05-19
Fix sellPowerup to disambiguate powerups by category at same slot

experintellia:
## Summary
Fixed a critical bug in `sellPowerup` where powerups from different categories (Upgrade, Ad, TeamMember) could occupy the same slot index, causing the wrong powerup to be removed. Also fixed a regression where the powerup grid would hang on a loading spinner after selling a powerup.

## Key Changes

**LocalEngine.ts (sellPowerup function):**
- Added logic to extract the game type from a powerup's `full_type` field to disambiguate powerups by category
- Changed the powerup lookup and removal filters from matching only `slot` to matching both `slot` AND `game_type`
- This ensures that when multiple powerups occupy slot 0 (one from each category), only the intended category's powerup is removed

**ProjectPerp.ts (SellPowerup handler):**
- Changed `gnode.data` reconstruction from `mergeData(groot.getTypeData(...), r.node?.instance_data)` to `mergeData(gnode.data, r.node?.instance_data)`
- This preserves runtime flags like `powerupsCached` that were being lost when rebuilding from type data
- Prevents the post-sell popup from getting stuck on the loading spinner with no powerup tabs

**Test Coverage:**
- Added comprehensive regression tests for cross-category slot-0 collisions, verifying that selling each category removes only the intended powerup
- Added test for error handling when attempting to sell a powerup category that doesn't exist at the specified slot
- Added e2e test to verify the popup grid renders correctly after selling a powerup (not stuck on spinner)

## Implementation Details
- Slot indices are per-category, so each category maintains its own 0-based grid
- The `_gameTypeFrom()` helper extracts the category from a powerup's `full_type` field (e.g., "UpgradePowerup:upgrade001" → "UpgradePowerup")
- The fix mirrors the existing slot-occupied guard in `buyPowerup` for consistency

https://claude.ai/code/session_01WHvLGLMAcajcjLp11vhuxp

### DIFF
--- a/scripts/LocalEngine.ts
+++ b/scripts/LocalEngine.ts
@@ -1302,10 +1302,28 @@ export function sellPowerup(
 
   var idata: NodeInstanceData = node.instance_data || {};
   var powerups: PowerupDef[] = idata.powerups || [];
+
+  // Slot indices are per-category (Upgrade / Ad / TeamMember each keep
+  // their own 0-based slot grid) but `powerups` is one flat array, so a
+  // bare `p.slot === slot` match resolves the wrong category's entry and
+  // the removal filter drops every entry sharing that slot index.  Scope
+  // the match to the sold powerup's game type — mirrors the buyPowerup
+  // slot-occupied guard above.
+  var soldDef = _findPowerupDef(perpTypeData, gestalt);
+  var soldGameType = soldDef ? _gameTypeFrom(soldDef.full_type) : '';
+  var _entryGameType = function (p: PowerupDef): string {
+    if (p.full_type) return _gameTypeFrom(p.full_type);
+    var d = p.gestalt ? _findPowerupDef(perpTypeData, p.gestalt) : null;
+    return d ? _gameTypeFrom(d.full_type) : '';
+  };
+  var _isSold = function (p: PowerupDef | null | undefined): boolean {
+    return !!p && p.slot === slot && _entryGameType(p) === soldGameType;
+  };
+
   var puEntry: PowerupDef | null = null;
   for (var i = 0; i < powerups.length; i++) {
     var p0 = powerups[i];
-    if (p0 && p0.slot === slot) {
+    if (p0 && _isSold(p0)) {
       puEntry = p0;
       break;
     }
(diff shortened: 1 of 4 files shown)


experintellia:
## Review — `sellPowerup` category disambiguation

Reviewed the engine fix and the `ProjectPerp` post-sell rebuild.

### Category disambiguation — correct and symmetric

`_gameTypeFrom` is robust (empty/undefined → `''`; no colon → whole string; with colon → prefix, so `"UpgradePowerup:upgrade001"` → `"UpgradePowerup"`). The key correctness property holds: **lookup and removal use the same `slot && _gameTypeFrom(full_type) === soldGameType` predicate**, so there's no "found but not removed" (or inverse) divergence — that asymmetry would have been worse than the original slot-only bug. `soldGameType` is derived from the caller-supplied gestalt via `_findPowerupDef`, not from a slot-only match, so no circular dependency. It mirrors the existing `buyPowerup` slot-occupied guard as claimed.

### `mergeData(gnode.data, …)` change — correct, consistent

`mergeData` is `Object.assign({}, type_data); Object.assign(data, instance_data)` with `instance_data` winning; the post-sell `instance_data.powerups` is the *complete* filtered array (not a delta), so no stale powerup entries survive. The switch makes `SellPowerup` consistent with `BuyPowerup`/`BuySlots` (already `mergeData(gnode.data, …)`) and preserves the `powerupsCached` runtime flag whose loss caused the stuck-spinner regression. Inverse-bug risk is low (type-level fields don't change on sell; `compilePowerups()` re-stamps immediately after) and `gnode.data` is guaranteed initialized (popup opened via `openPopup()`/`fetchPowerups()` before any sell).

### Test coverage — strong

All three categories seeded at slot 0 with distinct `full_type`s; each sell asserts only the intended gestalt is removed plus the exact `floor(price*0.75)` refund. The "category not at slot" path asserts `error:1` *and* that the unrelated entry is untouched (proves no fallback wipe). The e2e adds a real `fetchPowerups` baseline before asserting the grid re-renders (not stuck on `PopupContentLoading`), with a settle past the ~1.2s slot-swap animation so it can't pass on pre-rebuild DOM.

### Minor (pre-existing, out of scope)

If a malformed seeded entry has empty `full_type` *and* the sold gestalt isn't in `perpTypeData` (`soldGameType = ''`), `isSold` could match an unintended same-slot entry. In practice `buyPowerup` always stamps `full_type`, so this is a malformed-state edge, not a regression, and behavior is strictly better than the old slot-only filter. Not made worse here.

### Verdict

Approve — correct, symmetric (lookup/removal share one predicate), consistent with `buyPowerup`/`BuySlots`, and well-tested including the stuck-spinner e2e. The only residual is a pre-existing malformed-`full_type` edge that's out of scope.

---

PR #312, opened 2026-05-19
Port mission card to Preact, unify goal/reward view-model

experintellia:
## Summary

Ports the mission card (`views/mission.html`, `mission_goal_small.html`) from server-side templates to a Preact component (`MissionCard`), eliminating duplication of goal/reward computation logic between the card and the briefing/complete dialog (`MissionPopup`).

## Key Changes

- **New `MissionCard` component** (`scripts/components/MissionCard.tsx`): Renders the Missions-tab board card with title and goals, using the `small` variant of the shared `MissionGoalRow` component (no goal text, only sprite and progress).

- **Extracted `MissionGoalRow` component** (`scripts/components/popups/MissionPopup.tsx`): Refactored the inline goal rendering from `MissionPopup` into a reusable component that supports both full (with text) and small (card) variants via a `small` prop.

- **New `buildMissionCardProps` builder** (`scripts/game/missionView.ts`): Companion to `buildMissionPopupProps`, feeds the same `buildGoal` logic to ensure card and dialog render identical sprite/progress markup. Both consumers now share a single source of truth for goal/reward view-models.

- **Updated `RenderMissionPerp`** (`scripts/render/RenderTopLevelUI.ts`): Replaced server-side template rendering with Preact mount; now calls `buildMissionCardProps` and renders `MissionCard` directly on the board node. Added `remove()` override to properly unmount Preact when the card is destroyed.

- **Removed legacy templates**: `views/mission.html`, `mission_goal.html`, `mission_goal_small.html`, `mission_rewards.html` — their logic is now in `missionView.ts` and the Preact components.

- **Added parity test** (`tests/render/mission-card-dialog-parity.test.ts`): Renders the same fixture goal through both `MissionCard` and `MissionPopup`, asserting that the shared sprite/progress markup is byte-identical. Prevents silent re-forking of goal logic in future changes.

## Notable Details

- The per-goal view-model (`MissionGoalVM`) is computed once in `buildGoal` and reused by both card and dialog, eliminating the risk of ruleset/i18n changes drifting between the two render paths.
- The card renders *all* goals (no briefing 3-goal clip), matching the legacy `mission.html` behavior.
- Title HTML is rendered unescaped (legacy behavior preserved via `dangerouslySetInnerHTML`).
- Preact ownership of the card's DOM node is properly managed: no `jq.empty()` on re-render (diffs in place), and cleanup via `preactRender(null, el)` in `remove()`.

https://claude.ai/code/session_011FexbYpUCcVmqiSGAU76XX


experintellia:
## Review — mission card port + unified goal/reward view-model

Diffed `MissionCard` / `MissionGoalRow` against the deleted `views/mission.html` + `mission_goal_small.html`, and checked the dedup, the `RenderMissionPerp` mount/unmount, and the parity test.

### Does what it should — verified faithful

- **DOM equivalence vs the deleted templates: exact.** `mission.html` → `.MissionInline > .MissionLabel (unescaped title) + per-goal small rows`; the commented-out `MissionWrap` block was correctly *not* ported. `mission_goal_small.html` → `.MissionGoal.small[.complete] > .MissionGoalSprite + .MissionGoalProgress + .MissionGoalStatus` (no `.MissionGoalText`). `MissionCard`/`MissionGoalRow small` reproduce both, including the unescaped title (`dangerouslySetInnerHTML`, matching legacy `print(data.title)`) and the `MissionGoal small complete` class order. Data source matches too (`gameNode.data` ← legacy `D.gameNode.data`).
- **Dedup is real and the shared VM is the single source.** `buildGoal` is now fed by both `buildMissionPopupProps` and `buildMissionCardProps`; `MissionGoalRow` lives in a neutral `scripts/components/MissionGoal.tsx` (not under `popups/`, so the card doesn't reach into a dialog) — exactly the structure to avoid drift.
- **Unmount discipline correct.** `RenderMissionPerp.remove()` calls `preactRender(null, el)` before `super.remove()`, and `render()` deliberately drops the legacy `jq.empty()` (commented why — Preact diffs in place). This closes the unmount gap pattern from the #280-era review.
- **Stale `missionView.ts:1-12` comment fixed** (the inaccurate "collapses in tier 8" note is gone).
- The two bonus fixes are sound and self-guarded: the `.MissionGoal.small` whitespace-gap regression (Preact strips the inter-element whitespace the legacy `_.each` loop relied on for inline-block spacing — exactly the JSX-whitespace subtlety worth catching) and the `#GameContainer.DialogLock` menu lockout (the tab nav is a sibling above the Stage backdrop). Both have e2e guards.

### Findings

**1. The parity test's sprite-markup assertion is vacuous (low severity, but it reads stronger than it is).** The fixture has no `popup_sprite`, so `renderSpriteHtml` returns `''` and `spriteFragment` is the *empty* `<div class="MissionGoalSprite"></div>`. The "renders byte-identical sprite markup" assertion therefore only proves both sides render an empty sprite div — it does not exercise real sprite output. The actual anti-drift guarantee comes from `expect(cardProps.goals).toEqual(popupProps.goals)` (that *does* compare `spriteConfig`), so the intent is met — but the rendered-sprite test should use a fixture with a `popup_sprite` (or assert non-empty) so it guards what it claims. Also note the fixture deliberately avoids every `TEXT_*` sprintf branch (vendor `sprintf` absent under vitest) — fine for the dedup contract, but the parity test covers none of `buildGoal`'s text branches; the `toEqual` is carrying all the weight.

**2. Verify `RenderMissionPerp.render()` call frequency.** Each call rebuilds the full VM (`buildMissionCardProps` → `buildGoal` × all goals → `getType`/`getTypeData`/`sprintf`/`renderSpriteHtml`) and re-renders Preact with fresh prop identities (so the component always re-renders, never bails on equal props). Legacy did `jq.empty()+renderView` every call too, so this isn't a regression — but if `render()` is tick/frame-driven this is avoidable per-frame work the migration could have cut (render-on-data-change). Worth confirming it's state-change-driven, not per-animation-frame.

**3. Confirm the `.MissionPerp` node is Preact-content-only.** `preactRender(h(MissionCard,…), el)` makes Preact own `el`'s children. Legacy `jq.empty()` made the same assumption, so it's likely fine — but verify nothing else (a sprite/frame layer from the `RenderNode` base, or `draw()`) writes children into the same `jdomelem`, since Preact's diff will now reclaim anything it doesn't own.

**4. Minor:** `.MissionGoal.small { margin: 4px }` is an *approximation* of the legacy inline-block inter-element whitespace (which was font-size-dependent, horizontal + line-break, not a uniform 4px box) — the e2e only asserts `margin > 0`, not parity with master. Acceptable for a board card, just not pixel-identical. And the `#GameContainer.DialogLock .MainMenu` CSS hardcodes `#GameContainer` while `dialogManager` derives the node via `setup.renderContainer` — if that selector ever changes the menu-lock silently breaks; a one-line note tying them together would help.

### Verdict

Clean, faithful port that achieves the dedup it set out to (real shared VM, `toEqual` guard). No blockers. Fix #1 (give the parity fixture a sprite so the markup assertion isn't vacuous), and confirm #2/#3 (render frequency, node ownership) — both verifications rather than known bugs. The whitespace + dialog-lock follow-ups are good catches.

---

PR #313, opened 2026-05-20
Fix buy_powerup mission goals not completing for already-owned upgrades

experintellia:
## Summary

When a mission with `buy_powerup` goals seeded while the player already owned the target upgrades (e.g. "Gewinnspiel erweitern" requiring Logo & Design / Verteilen in Einkaufszentren / Praktikant), the objectives stayed at `0/1` until the player **sold and re-bought** each upgrade — only a fresh `buyPowerup` event advanced them.

The codebase already had `_autoCompleteBuyPerpGoals` to repair the equivalent problem for `buy_perp` goals (whose targets are standalone nodes). There was no sibling for `buy_powerup`, whose targets live in `node.instance_data.powerups[]`.

- Add `_autoCompleteBuyPowerupGoals(goals, nodes)` in `scripts/LocalEngine.ts` — same same-reference contract as the perp helper.
- Wire it into both spots the perp version runs: `_seedMissionGoals` (fresh mission / game load) and the cascade in `_completeMissionsIfReady` (mission unlocked by completing a prereq).
- Regression tests in `tests/handlers/handlers.test.js` covering the owned and not-owned cases. Verified the new test fails on the prior code and passes with the fix.

Existing players with currently-stuck objectives will see them clear automatically on next load, since the repair also runs at load time.

## Audit of other workflow types

Checked all seven workflows for the same class of bug:

| Workflow | Trigger | Repair needed? |
|---|---|---|
| `buy_perp` | event | already repaired |
| `buy_powerup` | event | **fixed here** |
| `integrate_profiles` | event, but computes against absolute state via `Math.max` | self-heals on next integrate event |
| `charge_perp`, `collect_cash`, `collect_profiles`, `upgrade_token` | cumulative event counters / per-collect — no persistent "already done" state | not vulnerable |

## Test plan

- [x] `pnpm test` — 721/721 passing
- [x] `pnpm lint` (biome) clean on changed files
- [x] `pnpm typecheck` clean on `scripts/LocalEngine.ts` (pre-existing unrelated errors elsewhere)
- [x] New regression test fails without the fix, passes with it
- [ ] Manual: load a save where mission006 is active and the upgrades are already owned — objectives should appear checked off

https://claude.ai/code/session_01QSvFUaofsYZZHNAP9xNFiU

---
_Generated by [Claude Code](https://claude.ai/code/session_01QSvFUaofsYZZHNAP9xNFiU)_

### DIFF
--- a/scripts/LocalEngine.ts
+++ b/scripts/LocalEngine.ts
@@ -1716,6 +1716,32 @@ function _autoCompleteBuyPerpGoals(
   return changed ? result : goals;
 }
 
+// buy_powerup analogue of _autoCompleteBuyPerpGoals. Powerups live in each
+// node's instance_data.powerups[] rather than as standalone nodes, so a
+// mission that unlocks after the player already owns the powerup would stay
+// stuck (the goal only advances on a fresh buyPowerup event — which is why
+// re-buying after a sell was the only workaround). Same same-reference
+// contract so callers can detect "nothing changed".
+function _autoCompleteBuyPowerupGoals(
+  goals: MissionGoal[],
+  nodes: GameNode[] | undefined
+): MissionGoal[] {
+  var owned: Record<string, true> = {};
+  (nodes || []).forEach(function (n) {
+    var pus = ((n.instance_data as NodeInstanceData) || {}).powerups || [];
+    pus.forEach(function (pu) {
+      if (pu && pu.gestalt) owned[pu.gestalt] = true;
+    });
+  });
+  var changed = false;
+  var result = goals.map(function (g) {
+    if (g.workflow !== 'buy_powerup' || g.complete || !owned[g.target]) return g;
+    changed = true;
+    return _completeGoal(g);
+  });
+  return changed ? result : goals;
+}
+
 function _eachMissionDef(ruleset: Ruleset, fn: (def: MissionDef) => void): void {
   if (!ruleset.missions) return;
   for (var i = 0; i < ruleset.missions.length; i++) {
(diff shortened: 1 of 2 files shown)


experintellia:
## Review — `_autoCompleteBuyPowerupGoals`

Faithful mirror of the existing `_autoCompleteBuyPerpGoals`. The pattern works.

### Correctness

- Same-reference contract preserved: early exit if no candidate `buy_powerup && !complete` goal, and `return changed ? result : goals` matches the perp sibling so callers can still detect "nothing changed". ✓
- Owned-set built once across all `nodes[i].instance_data.powerups[]`, indexed by `pu.gestalt`, with `pu && pu.gestalt` guarding malformed entries. ✓
- Goal predicate uses `g.target` (the powerup gestalt) — consistent with how `buyPowerup` advances the goal. ✓
- Wired in **both** spots the perp helper runs (`_seedMissionGoals` and the cascade in `_completeMissionsIfReady`). Order vs `_autoCompleteBuyPerpGoals` doesn't matter (independent workflows). ✓
- Runs on every `loadGame()`, so existing players with currently-stuck objectives clear on next load as the PR claims. ✓

### Workflow audit holds

Re-checked the table: `integrate_profiles` uses `Math.max` against the absolute snapshot so the next event/recheck self-heals; `charge_perp` / `collect_cash` / `collect_profiles` / `upgrade_token` are cumulative event counters with no persistent "already done" state to recover from. `buy_perp` / `buy_powerup` are presence-based + event-triggered → both vulnerable, both now repaired. Full 7-workflow coverage.

### One thing worth confirming

Like its perp sibling, this helper marks the goal complete **on presence**, regardless of `g.amount`. Powerup slots are per-category and per-gestalt, so a player normally owns ≤ 1 of each `target` gestalt at a time, and the test fixture uses `amount: null`. But if any ruleset ever defines a `buy_powerup` goal with `amount > 1` (intended as "buy this powerup N separate times across sell/rebuy cycles"), the helper would short-circuit to complete on a single owned copy and skip the counter. Worth confirming no ruleset goal has `amount > 1` for `buy_powerup`, or — if that's even a meaningful objective shape — that the perp helper's same over-complete-on-presence behavior is the accepted convention. The PR notes the new helper is the "Same same-reference contract" as the perp one, so I'd assume the answer is the latter (mirror established behavior), but flagging because the test only exercises `amount: null` / `amount: 1`-equivalent cases.

### Tests

Strong: one positive case (owned at load → completes), one negative (not owned → stays). Both drive through `loadGame()`, so they exercise the full `_seedMissionGoals` path rather than calling the helper directly. The PR's "fails on prior code, passes with fix" verification is the right gate.

### Verdict

Approve. Faithful mirror of the perp helper, both load and cascade wired, audit table is complete, tests are pointed. Only note: confirm no ruleset has a `buy_powerup` goal with `amount > 1` (or that presence-completes is the accepted convention as in the perp sibling).

---

PR #314, opened 2026-05-20
Show peer avatars in the highscore list

experintellia:
## Summary

Render the messenger's contact avatar between the rank number and the display name in each highscore row, behind the experimental webxdc avatar API proposed in [chatmail/core#6429](https://github.com/chatmail/core/pull/6429) + [deltachat-desktop#4481](https://github.com/deltachat/deltachat-desktop/pull/4481).

Per that PR, the avatar `user_id` is "basically the (self)addr" — so we already have everything we need:
- `webxdc.selfAddr` for the local user
- `delta.addr` for any peer we've received an update from

No new field on `Delta`, no member-list resolver, no extra plumbing.

## How it works

1. **Boot probe** — `boot.ts` fires a fire-and-forget `GET __webxdc__/avatar/<selfAddr>.jpg` once at startup. The result is cached in `webxdc-avatars.ts` as `_support: 'unknown' | 'supported' | 'unsupported'`. GET (not HEAD) because the avatar is served locally and HEAD support is uneven across webxdc implementations.
2. **Row decoration** — `getRanking()` checks `isAvatarSupported()` once per call, then decorates each `RankingRow` with `avatar: '__webxdc__/avatar/<addr>.jpg'` when supported. Rows on unsupported clients have no `avatar` field at all.
3. **Template** — `views/topscore_list.html` emits the `` only when `score.avatar` is set, and adds a `withAvatar` class to widen the row's padding. On unsupported clients the row layout is byte-identical to before, so there is no late layout shift from a 404.
4. **CSS** — circular slot (`border-radius: 50%`), 20px on regular rows, 26px on the self/user row.

## Test plan
- [x] `pnpm test` — 729/729 (incl. 5 new tests in `tests/handlers/avatars.test.js`)
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — clean
- [x] Manual smoke on a real Delta Chat client that ships the experimental avatar API (the dev simulator's `@webxdc/vite-plugins` doesn't emulate `__webxdc__/avatar/*`, so the probe stays `unsupported` in dev — by design, the highscore list looks identical to before).
- [ ] Verify no layout shift on a client that does ship avatars but where a specific peer has no picture (per chatmail/core#6429 the URL 404s for that peer — the probe still succeeded for self, so the slot is reserved; the broken-image fallback handling on a per-row basis is a separate concern if it arises).

---
_Generated by [Claude Code](https://claude.ai/code/session_0145jMCpqJC9r3qg25PDHpUZ)_

### DIFF
--- a/css/Render.css
+++ b/css/Render.css
@@ -1251,9 +1251,11 @@
   display: block;
   object-fit: cover;
 }
-/* Placeholder: the in-game Profiles token icon, centred in the slot. The
-   <img> sits over it; on a 404 the <img> removes itself so this shows. */
-.TopscoreList.hasAvatars .TopscoreAvatar::before {
+/* Placeholder: the in-game Profiles token icon, scaled up to fill the
+   slot. Rendered only after the row's <img> has hit onerror and added
+   `.failed` — so successful avatars don't briefly flash the placeholder
+   while they're loading. */
+.TopscoreList.hasAvatars .TopscoreAvatar.failed::before {
   content: '';
   position: absolute;
   top: 5px;
@@ -1261,6 +1263,8 @@
   width: 16px;
   height: 16px;
   background: url(../img/MainSprites.png) no-repeat -436px -616px;
+  transform: scale(1.5);
+  transform-origin: center;
 }
 
 .TopscoreDisplayname {
--- a/css/Render.css
+++ b/css/Render.css
@@ -1211,7 +1211,6 @@
   border-radius: 6px;
   background: #FFF;
   height: 32px;
-  width: 542px;
   box-sizing: border-box;
   padding: 0 6px 0 4px;
   display: flex;
@@ -3366,10 +3365,6 @@ button {
     max-width: 92vw;
     box-sizing: border-box;
     box-shadow: 3px 3px 0px #aaa, 3px 5px 8px #000;
-  }
-  .TopscoreRow {
-    width: auto;
-    max-width: 100%;
   }
   .TopscoreRankBarWrap {
     max-width: 100%;
--- a/css/Render.css
+++ b/css/Render.css
@@ -1214,7 +1214,7 @@
   height: 24px;
   width: 542px;
   display: inline-block;
-  padding-left: 60px;
+  padding-left: 32px;
 }
 
 .TopscoreRow.user {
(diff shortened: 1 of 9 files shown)


experintellia:
## Review — highscore avatars

Tight, well-scoped PR. The architecture — probe once at boot, gate row decoration on the resolved support flag — is the right shape: it cleanly avoids the late-404 layout-shift trap a naïve "always render ``" would have. The `_support: 'unknown' | 'supported' | 'unsupported'` tri-state defaulting `isAvatarSupported()` to `false` until resolution means the row class is set deterministically at template render time. Good test shape (unit-test the decoration gate; document why the network path isn't unit-tested).

### Approve, with a few verify-on-real-client items

**1. Per-peer 404 fallback** (`__webxdc__/avatar/<addr>.jpg` 404s when a contact has no picture per chatmail/core#6429)
With the slot reserved at the row level (probe says "supported"), an individual peer without a picture will leave a broken `` in the row. The `background: #ddd` on `.TopscoreAvatar` shows a gray placeholder behind the broken image, but most browsers also overlay a broken-image glyph on the foreground. Two cheap mitigations to consider in a follow-up if it looks bad in practice:
- Add `onerror="this.style.visibility='hidden'"` (or `display:none`) on the `` in `topscore_list.html` — keeps the gray circle, hides the glyph.
- Or skip the `` element altogether and rely purely on the CSS background — but that loses the `object-fit: cover` cropping when the image does load.

Not a blocker; you flagged this exact case in the test plan.

**2. URL encoding (`@` → `%40`)** — `getAvatarUrl('alice@test')` produces `__webxdc__/avatar/alice%40test.jpg`. The chatmail/core#6429 spec says the user_id "basically is the (self)addr" but doesn't (last I checked) commit to either encoded or raw form. The probe will catch this — if the messenger expects raw `@` and 404s on `%40`, the probe fails and we silently fall back to the no-avatar layout. So it's self-healing if wrong, just worth a real-client smoke to confirm the encoding actually matches.

**3. First-leaderboard race** — `boot.ts` fires `probeAvatarSupport(selfAddr)` fire-and-forget. If the user opens the leaderboard within the few hundred ms before the probe resolves, rows render without avatars on a supporting client; subsequent renders include them. Not a layout shift (the row class flips on next render), but a one-time "where are the pictures?" cosmetic blip. Acceptable trade-off vs. blocking boot on the probe — but if you want to close the race, `await probeAvatarSupport(selfAddr)` in boot would do it at the cost of one local fetch in series.

### Nits

- `topscore_list.html`: `score.avatar || ''` then `avatar_url ? ' withAvatar' : ''` works but doubles the truthiness check. Could compress to `var has_avatar = !!score.avatar;` and use that in both spots. Cosmetic.
- The `_support` module-level state is process-global; the `__resetAvatarsForTest` / `__setAvatarSupportForTest` test backdoors are the right pattern for this. Good.
- `webxdc-avatars.ts` is the first new module to use `const`/`let` instead of `var`. The existing codebase uses `var` more often — not a blocker, but worth a heads-up that the style is mixed now.

Good shipping shape; the probe-and-gate pattern is well-considered.

---

PR #315, opened 2026-05-20
Mobile dialog redesign for iPhone SE — Tutorial / Mission / Contact / Status + visual regression suite

experintellia:
Iterative pass through the mobile dialog inventory, targeting iPhone SE gen 2/3 (375×667) as the minimum supported viewport.  Each task below was driven by user feedback and is its own commit so the PR can be reviewed dialog-by-dialog.

## What's in
| Commit | Task | Summary |
|--------|------|---------|
| `f6e53fc` | Tutorial / LevelUp width | `.PopupBody.TutorialBody` was hard-pinned 588 px and overflowed the 375 px viewport.  Shrink to `calc(100vw - 16px)` with right-anchored inner text; covers LevelUp too via shared `.TutorialBody`. |
| `d7bd2c1` | Symmetric margins | `.Popup { width: 600 }` + 6 px margin left a 4 px strip on the right.  Auto-size the wrapper to the body and pull `.Tutorial .NotificationBubble` flush so flex centring is symmetric. |
| `33e5be6` | Mission width + statusbar clearance | Same 588 / 600 / 440 px design as Tutorial, plus a rotated 600 px `.MissionDecorator`.  Reflow all fixed-width children; `margin-top: 24px` so the decorator's negative top + 1° rotation overhang clears the statusbar. |
| `6787827` | C1-C3 + F | (C1) Drop forced-grow `min-height`; content-sized natural dialog. (C2) `.MissionRewards` becomes a wrapping flex row with "Rewards" title on its own line. (C3) Inner-scroll inside `.MissionContent` activates when content exceeds `max-height: calc(100vh - 24px)` (verified with mission006 3-goal/2-reward stress test). (F) Adds `tests/e2e/dialog-screenshots.spec.ts` — `toHaveScreenshot` baselines for Tutorial + Mission at iPhone SE 375×667 and desktop 1280×800.  Workflow on deliberate visual changes: `pnpm test:e2e tests/e2e/dialog-screenshots.spec.ts --update-snapshots`. |
| `3898a22` | Scroll boundary | Move the scroll container from `.PopupBody` to `.MissionContent` so the close X (`top: -6; right: -6`) and the `.MissionDecorator` banner (-14 px) aren't clipped by the body's `overflow: hidden`.  `.PopupTitle` and `.NotificationBubble` become fixed flex children; only `.MissionWrap` scrolls. |
| `2aa9a91` | C4 + C5 + G | (C4) Hide `0/1` progress on single-unit goals via `missionView.ts`; the checkmark conveys the state.  Lift `.MissionGoal { height: 58px }` → dynamic min-height; vertically centre the status sprite via `top: 50% / translateY(-50%)`.  (C5) Restore variant-specific paddings on `.MissionReward.<type>` (was clobbered).  (G) Hide the @webxdc/vite-plugins "webxdc dev tools" panel in snapshots via inline-style selector. |
| `12ce4bc` | D-1, D-2, E | (D-1) Raise `.PopupContainer.lockOn` z-index 100 → 20000 so it sits above the in-Stage `.Statusbar` (z=10000) — fixes status indicators painting over the perp-popup header on phone viewports and making the X "unclickable".  (D-2) Generic `@media (max-width: 768px)` clause for the bare `.Popup` / `.PopupBody` — covers Contact / Client / Pusher / Token / Karma / About in one block.  (E) `e.preventDefault()` on the `.StatusItem` touchend handler so the synthesised compatibility click never reaches the just-mounted popup container; status popups stay open after a real touch tap.  Includes a mobile-touch regression test using a `hasTouch` context. |
| `8ce99b6` | C7-C10 + Contact fixture | (C7) Revert mission goal to original block layout so `.MissionGoalProgress` flows below the goal copy where it always did; still vertically-centre the status stamp on dynamic-height cards.  (C8) `min-height: 58` matches desktop's hardcoded height; tighter `.MissionGoalText` padding.  (C9) Rehouse reward icons inside the chip (was overhanging via `left: -33` etc.); bump padding-left to 58 px and chip min-height to 48 px so Cash / Profiles / Risk / XP / Penalty all fit cleanly with the flex gap absorbing inter-chip spacing.  (C10) Floating OK button at the bottom — `position: absolute; bottom: 12` outside the scroll wrapper; speech bubble and scroll wrapper get padding-bottom so scrolled content never tucks beneath it.  Contact screenshot fixture: enrich `data.ProfileSet.tokens_set` with `getTypeData(gestalt)` per token so per-token sprites + labels render the same way as a real canvas tap. |

## Skipped intentionally
- **C6** (let the Mission popup cover the menu tabs on mobile) — touches stacking contexts between `#GameContainer`'s direct children.  The popup's container lives inside the Stage and the MainMenu is a sibling; making the popup cover the menu needs either `position: fixed` escape on the lockOn container or a Stage z-index lift coordinated with the dialog manager.  Worth its own PR.
- **B** (Imperium camera-pan centring on narrow viewports) — the database-tutorial step pans to a (1024, 840) viewmap position but the Database perp ends up at viewport x ≈ -68 px on iPhone SE.  Root cause is `RenderViewMap.scrollTo` (`scripts/render/RenderViews.ts:734-745`) mixing unscaled-content coords with viewport-pixel coords; the Zynga scroller interprets the result in scaled-content space (zoom 0.75 on Imperium).  Fix is JS-only — scale `pos` by `this.zoomScale` before subtracting `vpCenter`, clamp against the scroller's actual max, mirror in `GameRoot._centerActiveView`.  Defer to a follow-up PR so this one stays scoped to dialog layout.

## Test coverage
- `tests/e2e/dialog-lifecycle.spec.ts` — 49 existing tests still pass (one-off Section P flake recovered on retry).
- `tests/e2e/mobile-dialog-fit.spec.ts` — three pin-this-doesn't-overflow tests for Tutorial, Mission and a `hasTouch` regression for status popups stay-open-after-real-tap.
- `tests/e2e/dialog-screenshots.spec.ts` — 10 baselines (5 dialogs × {mobile, desktop}) — Tutorial, mission001, mission006, Cash status, Contact (Jessica) — committed under `tests/e2e/dialog-screenshots.spec.ts-snapshots/`.

## Touch points
- `css/Render.css` — all mobile rules live inside `@media (max-width: 768px)` (single base-rule change: `.PopupContainer.lockOn z-index 100 → 20000`).
- `scripts/components/popups/MissionPopup.tsx` — only conditional render of `.MissionGoalProgress` when non-empty.
- `scripts/game/missionView.ts` — `progressHtml` omitted for single-unit goals.
- `scripts/render/RenderTopLevelUI.ts` — `e.preventDefault()` on `.StatusItem` touchend (E fix).

https://claude.ai/code/session_018n8Y4s4K66g9oDjAUnZW15

---
_Generated by [Claude Code](https://claude.ai/code/session_018n8Y4s4K66g9oDjAUnZW15)_


experintellia:
## Review — mobile dialog redesign for iPhone SE + visual regression suite

Substantial, well-iterated pass. The commit-per-task structure makes the diff legible dialog-by-dialog, and the standing rule (all mobile changes live inside `@media (max-width: 768px)`, with one carefully-justified base-rule lift on `.PopupContainer.lockOn` z-index) keeps the blast radius narrow. The new visual regression suite + the `mobile-dialog-fit.spec.ts` bounding-box assertions are the right combination: the screenshots catch unintended desktop drift, the box-fit tests pin specific invariants that snapshots can't (e.g. "the body's right edge stays inside the viewport"). 49 lifecycle tests still passing is the cherry on top.

I verified two specific structural concerns:
- **`.PopupBody` is `position: relative`** (Render.css:1422), so the floating mobile OK button (`.NotificationButtons { position: absolute; bottom: 12px }`) anchors to the popup body, not the viewport-sized overlay. ✅
- **`.MissionContent` is `position: static` on desktop** (Render.css:2239), so the mobile `flex: 1 1 auto` is not silently ignored as a non-flex-item rule. ✅

### Approve, with a few items to consider

**1. D-2 (generic `.Popup` / `.PopupBody` mobile shrink) is implicitly the biggest blast radius in this PR.**
```css
.Popup { width: auto; margin: 0; }
.PopupBody { width: calc(100vw - 12px); max-width: 588px; ... }
```
This affects *every* popup that doesn't have a more specific override: Contact, Client, Pusher, Proxy, Token, Karma, About, Highscore. The screenshot suite covers Status-Cash + Contact (Jessica); the other six aren't pinned. Given the recent Preact tier-5 migrations, a few of these (especially Karma + About, which have unusual internal absolute-positioned children) could subtly misfit at 375 px without your noticing. **Consider adding one screenshot baseline each for About + Karma + Client (Imperium)** in a follow-up so the safety net catches the full perp-popup family. Not a blocker for this PR.

**2. `MissionGoalProgress` is now conditionally absent from the DOM.**
`missionView.ts` returns `progressHtml: ''` for single-unit goals (`amount <= 1`), and `MissionPopup.tsx` renders `{pr && (...)}`. So the `<div class="MissionGoalProgress">` is omitted entirely — not just empty. This is a DOM-structure change from the legacy template (which always emitted the div). Worth confirming:
- No CSS rule depends on `.MissionGoalProgress` being present for height/spacing of `.MissionGoal` (I scanned the rules around line 1036–1104; they're all `.MissionGoalProgress { ... }` self-contained, so this looks safe).
- No goal-progress visibility test asserts on the empty `0 / 1` text rendering. (`dialog-lifecycle.spec.ts` doesn't seem to — confirmed by your "49 still pass" note.)

I think you're clear here; just naming the change so future debugging knows where to look.

**3. The webxdc-vite-plugins panel hider is brittle:**
```ts
'body > div[style*="z-index: 9999"][style*="position: fixed"] { display: none !important; }'
```
This depends on the *exact* inline-style format emitted by `@webxdc/vite-plugins`. If the plugin ever switches from inline style to a class or changes the z-index, this silently fails and the dev-tools panel leaks into baseline snapshots — and the next `--update-snapshots` re-bakes the leak. Two cheap upgrades:
- Add a defensive `data-testid` attribute selector if the plugin emits one (worth checking).
- Or assert at test-time that the panel is actually hidden post-`addStyleTag` (`await expect(panel).toBeHidden()`) so a silently-failed hide turns into a test failure, not a corrupted snapshot.

**4. The screenshot suite's drain loops are duplicated** between `dialog-screenshots.spec.ts:50-71` and `mobile-dialog-fit.spec.ts:32-67` (and the contact-test re-drains after `page.reload()`). Up to 40 iterations × 300ms = 12 seconds per test in the worst case. With 10 screenshot baselines + 3 mobile-fit tests, the suite-level worst case is substantial. Consider extracting `drainBootPopups` to a shared helper (e.g. `tests/e2e/_dialog-helpers.ts`) so a future fix to the drain logic doesn't have to land in two places. Cosmetic, not blocking.

**5. `e.preventDefault()` on `.StatusItem` touchend** — the comment explaining the synthesised-click → backdrop-tap chain is exemplary. The fix is correctly scoped (`if (e.type === 'touchend')` so desktop click is untouched), and you wired up a real `hasTouch: true` context to exercise it. Solid.

**6. `.PopupContainer.lockOn z-index 100 → 20000`** — the inline comment correctly maps the stacking model (`.Statusbar` z=10000 below, `.PopupContainer.Top` z=100000 above). The one thing to double-check: any other in-Stage element with z-index in the [100, 20000) range that *was* covering the popup container before and is now covered by it. The big ones are the menu tabs and any toast variants; if you defer C6 (popup covers menu) to a follow-up, the menu tabs are below the popup at z=? Worth a glance during your manual smoke.

### Nits

- The MissionRewards reflow comment (~line 3467) is great — it explains *why* the desktop overhang pattern can't survive flex `gap`. Keep this kind of inline doc when reverting design intent.
- `B` (Imperium camera-pan) deferred with a JS-only proposed fix in the PR body — good capture; that's a separate problem worth its own PR.

Overall: ship it. The visual-regression baselines are now part of the contract — be aware that any deliberate dialog change will need `--update-snapshots` + a reviewer eyeballing the PNG diffs. That's the right trade-off for the level of mobile precision this PR is aiming at.

---

PR #316, opened 2026-05-20
Show energy refill timer without hover when AP is 0, and inside the AP dialog

experintellia:
## Problem

On mobile there is no hover, so the "More Energy in MM:SS" countdown on the statusbar was completely invisible when the player had no energy. The same information was also missing from the AP info dialog, which is the only thing a mobile user can tap to learn about their energy.

### Always-visible timer on the statusbar when AP is 0
`RenderTopLevelUI.ts`
- Added a separate `autoLoop` (independent of the hover `loop`) so the two timers don't interfere.
- `refreshAPRemainIfEmpty()` is called at the end of every `render()`. When `AP_val === 0` it immediately shows `.StatusRemain` and starts a 1 s tick loop — no mouse interaction required.
- `mouseleave` guard: when `ap_value === 0` the leave handler returns early, so a stray mouse-out doesn't hide the always-visible tooltip.

### Refill timer inside the AP info dialog
`APStatusPopup.tsx` + `GameRoot.ts`
- `APStatusPopup` gains an optional `apRemaining?: number` prop (ms until next tick).
- `GameRoot` computes the remaining time via `APTicker?.getRemainingTime?.()` and passes it when opening the popup.
- When `apValue < apMax` and `apRemaining` is set, the dialog renders a gold **"More Energy in HH:MM:SS"** line below the description.

`Render.css`
- `.PopupText.APRemain` — gold (`#ffd800`) text matching the statusbar tooltip colour.
- `.PopupText.APRemain .highlight` — white for the time value so it pops.

### Tests
`tests/e2e/energy-refill-timer-visibility.spec.ts` — 4 new e2e tests:
1. Timer appears on the statusbar **without hover** when AP drops to 0
2. Timer disappears again once AP is restored to full
3. AP info dialog **shows** the refill countdown when AP < max
4. AP info dialog shows **no** refill line when AP is full

`views/statusbar.html` — added `data-testid="dd-ap-remain"` to `.StatusRemain` so tests can target it without fragile CSS selectors.

https://claude.ai/code/session_013EK7YGGDKRNvNUHL15642h

---
_Generated by [Claude Code](https://claude.ai/code/session_013EK7YGGDKRNvNUHL15642h)_

### DIFF
--- a/css/Render.css
+++ b/css/Render.css
@@ -2431,6 +2431,13 @@
 .PopupBody.Status .PopupHeader {
   height:260px;
 }
+.PopupText.APRemain {
+  margin-top: 6px;
+  color: #ffd800;
+}
+.PopupText.APRemain .highlight {
+  color: #fff;
+}
 .PopupBody.TutorialBody {
   background: #FFF;
   box-shadow: 0px 3px 12px rgba(0,0,0,0.3);
--- a/css/Render.css
+++ b/css/Render.css
@@ -2431,12 +2431,39 @@
 .PopupBody.Status .PopupHeader {
   height:260px;
 }
-.PopupText.APRemain {
-  margin-top: 6px;
-  color: #ffd800;
+.PopupText.APRefillLabel {
+  margin-top: 8px;
+  color: #888;
+  font-size: 13px;
 }
-.PopupText.APRemain .highlight {
-  color: #fff;
+.APRefillBar {
+  position: relative;
+  margin: 4px 0 0;
+  height: 26px;
+  background: #d4d4d4;
+  border-radius: 13px;
+  overflow: hidden;
+}
+.APRefillBarFill {
+  position: absolute;
+  left: 0;
+  top: 0;
+  height: 100%;
+  background: linear-gradient(to right, #e07820, #ffd800);
+  border-radius: 13px;
+  transition: width 0.5s linear;
+}
+.APRefillBarText {
+  position: absolute;
+  inset: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-family: Voltaire;
+  font-size: 14px;
+  font-weight: bold;
+  color: #333;
+  text-shadow: 0 1px 2px rgba(255, 255, 255, 0.7);
 }
 .PopupBody.TutorialBody {
   background: #FFF;
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,14 +8,18 @@
       "name": "data-dealer",
       "version": "0.2.0",
       "license": "GPL-3.0-or-later",
+      "dependencies": {
+        "preact": "^10.29.1"
+      },
       "devDependencies": {
         "@biomejs/biome": "1.9.4",
         "@playwright/test": "^1.56.1",
         "@types/node": "^22.0.0",
         "@vitest/coverage-v8": "^3.0.0",
         "@webxdc/types": "^2.1.2",
-        "@webxdc/vite-plugins": "^1.6.0",
+        "@webxdc/vite-plugins": "github:experintellia/vite-plugins#claude/indexeddb-webxdc-simulator-Oy3ii",
         "lefthook": "^2.1.6",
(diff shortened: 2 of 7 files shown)


experintellia:
## Review — energy refill timer (statusbar always-on + AP dialog)

Genuine mobile UX problem identified and fixed: without hover, mobile players had no visibility into AP refill timing. The two-pronged fix (always-on statusbar tooltip when AP=0; static refill line inside the AP info dialog) is the right scope. Code is straightforward and the e2e coverage exercises both surfaces. Approve with a few items to consider.

### Findings

**1. The dialog timer is frozen at open-time. (UX gap)**
```tsx
<span class="highlight">{toTime(apRemaining)}</span>
```
`apRemaining` is computed once in `GameRoot.click_status.AP` and passed as a prop snapshot. If a user opens the dialog at 3:47 remaining and reads the description for 30 seconds, the dialog still shows 3:47 — no countdown. The statusbar timer next to it auto-updates every second, which makes the dialog one look broken by comparison. Two cheap fixes:
- Add a 1s ticker inside `APStatusPopup` that re-reads `groot.APTicker.getRemainingTime()` via a callback prop, or
- Pass a `getRemaining: () => number` prop and useEffect a 1s interval to re-render `toTime(getRemaining())`.

Not a merge blocker — current behavior is strictly better than no info — but worth a follow-up tick.

**2. State-source inconsistency between the two guards.**
```ts
// refreshAPRemainIfEmpty (render layer):
if (this.AP_val > 0) return;
// mouseleave handler (engine state):
if (groot && groot.ap_value === 0) return;
```
The two checks reference different sources of truth: `this.AP_val` is the render-layer cache (populated during `render()`), `groot.ap_value` is the engine state. They're usually in sync, but if a tick fires `ap_value=1` between the render call and the mouseleave handler, the autoLoop is stopped (because `AP_val` is still 0 from the last render) while the mouseleave guard hides the tooltip (because `ap_value === 0` is false). End result: tooltip stays hidden until the next render, then re-appears. Minor flicker possibility. Recommend using `groot.ap_value` in both places for consistency.

**3. `mouseleave` early-return can leave the hover `loop` running.**
When AP=0 and the user mouses over the statusbar and then off it:
- `mouseenter` (existing handler, unchanged) starts the hover `loop`.
- `mouseleave` returns early — `stopLoop()` is never called.
- Now both `loop` (hover) and `autoLoop` (always-on) are firing at 1 Hz, both writing the same content to `.StatusRemain`.

Functionally correct (both produce the same text) but doubled work. The cleanest fix is to `stopLoop()` in the mouseleave handler even when AP=0 — the autoLoop already keeps the tooltip alive, so the hover loop is redundant.

**4. Unused `$` binding.**
```ts
const $ = getRenderJQuery('RenderTopLevelUI');  // never used in this function
```
Drop it; `jq.find()` already covers it.

**5. Test access-pattern mix.**
Tests interleave `(window as any).require('app').getApplication().game` and `(window as any).__dd._app.game` to reach the same `GameRoot`. The `__dd` accessor is the canonical test backdoor in `dialog-lifecycle.spec.ts` — recommend unifying on it for consistency, makes future test-helper extraction easier.

**6. `data-testid="dd-ap-remain"` on the still-EJS-rendered `views/statusbar.html`.**
Fine for now — statusbar isn't on the Preact migration list. Just noting that when statusbar eventually ports to Preact, this testid needs to come along.

### Nits
- `startAutoLoop` self-recurses via `setTimeout` rather than `setInterval`. The existing `loop` pattern likely does the same — consistent with the codebase, fine.
- `+remaining` numeric coercion in `GameRoot.ts` is defensive but `getRemainingTime()` is typed as returning a number, so `apRemaining = remaining` would do.
- `.PopupText.APRemain` gold `#ffd800` on dark dialog body is readable; the `.highlight` white-on-gold border is a touch low contrast — verify on the actual `.PopupBody.Status` background colour during smoke.

Solid PR for the scope stated. The frozen dialog timer is the only real UX gap; everything else is polish.

---

PR #323, opened 2026-05-21
Fit the About popup to the viewport on mobile

experintellia:
## Summary

The About dialog is a single fixed-height (428 px) scroll area with the Close button pinned absolute at `bottom:-10px`. On a phone the prose is far taller than the box, so:

- mid-scroll text slides **behind** the Close button,
- the popup **overflows the viewport** (button pushed off-screen), and
- the long GitHub URL (one unbreakable token) spills past the right edge.

### Changes

- **`AboutPopup.tsx`** — `.PopupButtons` moves out of `.PopupContent` so the mobile layout can pin it as a static footer. On desktop it stays absolute-positioned relative to `.PopupBody`, so desktop rendering is unchanged.
- **`Render.css`** — mobile-only (`max-width:768px`) reflow scoped to `.PopupBody.About`: a flex column capped just short of the viewport. `.PopupContent` is the flexible scroll area; the Close button is a static footer that stays visible below it. `overflow-wrap` lets the URL break.
- The cap (`calc(100vh - 100px)`) reserves space for the reflowed mobile statusbar — the popup is flex-centred inside `.Stage` (viewport minus the ~84-px statusbar), not the raw viewport, so a `100vh`-only cap would still overflow.

## Test plan

- [x] Added a desktop+mobile screenshot test pair for the About popup
- [x] Verified at iPhone SE: popup fits, content scrolls internally, Close button always visible, no overlap, URL wraps
- [x] Desktop layout unchanged (media query + `.PopupBody.About` scope; `about-desktop` baseline)
- [x] `pnpm typecheck` + unit tests + biome pass

### Note for the reviewer

There is a parallel branch `claude/mobile-about-dialog` taking a different approach to the same dialog — this PR is meant to be compared against it.

The `dialog-screenshots` suite is `test.skip` on CI and host-tied by design (anti-aliasing differs per host); the `about-*` baselines here were generated in this environment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


experintellia:
IMO this pr is a regression compared to master, closing

PR #344, opened 2026-06-15
feat: game save export/import via webxdc sendToChat + importFiles (#127)

### LINKED ISSUE #127: Phase 9: game save export/import via webxdc sendToChat + import files API
## Idea

Allow players to export and import their game save using WebXDC's `sendToChat` API (for export) and the WebXDC import files API (for import). This would let players back up progress, transfer saves between devices, or share saves with others.

## Motivation

Players may want to preserve progress, switch devices, or start fresh in a new chat with an existing save. WebXDC already provides the primitives needed for this without requiring any server infrastructure.

### Export
- Add an "Export Save" option in the game UI (e.g. settings or profile screen).
- Serialize the current game state/profile to a file (e.g. JSON).
- Call `window.webxdc.sendToChat({ file: { name: "data_dealer_save.json", blob: ... } })` to let the player send the save file into any chat.

### Import
- Add an "Import Save" option that uses the WebXDC import files API to let the player pick a previously exported save file.
- Validate and load the save data, replacing the current local state.
- **After a successful import, automatically send a visible message to the chat** (via `window.webxdc.sendUpdate`) indicating that the player imported a profile — something like *"[Player] loaded an imported save."*
  - This keeps imports transparent to other participants and makes it reasonably easy to spot potential cheating (e.g. importing a highly progressed save mid-game).

## Anti-Cheat Transparency

The chat notification on import is intentional and important. It does not prevent cheating but makes it visible to all chat members, which is sufficient for the casual/social nature of the game.

## Notes

- **No priority** — this should only be considered after all current phases/waves are complete.
- File format versioning should be considered so that old saves can be migrated on import.
- The import should fail gracefully with a user-facing error if the file is malformed or from an incompatible version.

experintellia:
Closes #127.

Adds **Export Save** / **Import Save** to the About / settings popup.

## Export
- Serializes **only the current player's own progress** and hands it to `webxdc.sendToChat` as `data_dealer_save.json`.
- Per the request on the issue, the snapshot deliberately **excludes `state.peers`** (every *other* player's aggregated leaderboard stats) **and `state.addr`** — an export contains the current user's state only, never other users'. `buildSaveState` opt-ins each field explicitly so a future `LocalState` field can't silently leak into exports.

## Import
- Picks a file via `webxdc.importFiles({ extensions: ['.json'] })`, validates format + version, then persists an `importSave` delta.
- The reducer replaces the player's own progress while **preserving local identity (`addr`) and the peer table** — importing someone else's save can't hijack your address or wipe the leaderboard. Foreign `importSave` deltas are blocked by `applyDelta`'s existing addr guard.
- Because the delta is the source of truth, the import **survives cold-start replay** (mirrors how `setLocale` persists; the popup reloads on success).
- A visible chat notice — *"&lt;player&gt; loaded an imported save."* — rides on the same `sendUpdate` for the issue's anti-cheat transparency requirement. Top-level `game_values` is mirrored on the delta so the leaderboard row refreshes from the imported save.
- Malformed / newer-version files fail gracefully with a localized message; older `save_version`s forward-migrate via `freshState` defaults.

## Files
- `scripts/state.ts` — `SAVE_FORMAT`/`SAVE_VERSION`, `buildSaveState`/`buildSaveFile`/`parseSaveFile`, `importSave` reducer + op.
- `scripts/LocalEngine.ts` — `exportSave`/`importSave` handlers.
- `scripts/components/popups/AboutPopup.tsx` + `scripts/game/GameRoot.ts` — Export/Import UI wired through `app.remote`.
- `scripts/app.ts` — typed `AppRemote` signatures.
- `i18n/{en_US,de_AT}.{json,po}` — new strings (en + de).
- `css/Render.css` — in-flow save section (doesn't collide with the pinned Close button).
- `tests/state/save-export-import.test.js` — round-trip, validation, reducer (current-user-only, identity/peer preservation, replay safety).

## Verification
- `pnpm typecheck`, `pnpm lint`, `pnpm test` (754 tests, incl. 15 new) all pass.

https://claude.ai/code/session_011BACZuQDuVPdVmy33JLomf

---
_Generated by [Claude Code](https://claude.ai/code/session_011BACZuQDuVPdVmy33JLomf)_

### DIFF
--- a/css/Render.css
+++ b/css/Render.css
@@ -1770,6 +1770,30 @@
 .PopupBody.About .PopupTab {
   padding-bottom: 56px;
 }
+/* Save export/import section (issue #127).  Unlike the base
+ * `.PopupButtons` (absolutely pinned to the dialog's bottom edge), these
+ * buttons live in-flow inside the scrollable `.PopupContent` so they sit
+ * with the surrounding prose and never overlap the pinned Close button. */
+.PopupBody.About .SaveSection {
+  text-align: center;
+}
+.PopupBody.About .SaveButtons {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  gap: 12px;
+  margin: 12px 0 4px;
+}
+.PopupBody.About .SaveButtons .Button {
+  position: relative;
+}
+.PopupBody.About .SaveButtons .Button.disabled {
+  opacity: 0.5;
+  pointer-events: none;
+}
+.PopupBody.About .SaveError {
+  color: #cd495d;
+}
 
 .SubpopContainer {
   position: absolute;
--- a/i18n/de_AT.json
+++ b/i18n/de_AT.json
@@ -980,5 +980,41 @@
   "achievement_karma_devil": [
     null,
     "%s hat den vollen Teufelsstatus erreicht"
+  ],
+  "save section headline": [
+    null,
+    "Spielstand"
+  ],
+  "save section text": [
+    null,
+    "Sichere deinen Fortschritt in einem Chat oder lade einen zuvor exportierten Spielstand. Beim Importieren wird dein aktueller Fortschritt ersetzt."
+  ],
+  "Export Save": [
+    null,
+    "Spielstand exportieren"
+  ],
+  "Import Save": [
+    null,
+    "Spielstand importieren"
+  ],
+  "save export chat text": [
+    null,
+    "Mein Data-Dealer-Spielstand"
+  ],
+  "save imported chat info": [
+    null,
+    "%s hat einen importierten Spielstand geladen."
+  ],
+  "save import error malformed": [
+    null,
+    "Diese Datei ist kein gültiger Data-Dealer-Spielstand."
+  ],
+  "save import error version": [
+    null,
+    "Dieser Spielstand wurde mit einer neueren Version des Spiels erstellt und kann nicht importiert werden."
+  ],
+  "save import error unavailable": [
+    null,
+    "Das Importieren von Spielständen wird hier nicht unterstützt."
   ]
 }
--- a/i18n/de_AT.po
+++ b/i18n/de_AT.po
@@ -1167,6 +1167,46 @@ msgstr "Kapazität für %s Projekte"
 msgid "per deal"
 msgstr "Füttern"
 
+#: scripts/components/popups/AboutPopup.tsx
+msgid "save section headline"
+msgstr "Spielstand"
+
+#: scripts/components/popups/AboutPopup.tsx
+msgid "save section text"
+msgstr ""
+"Sichere deinen Fortschritt in einem Chat oder lade einen zuvor exportierten "
+"Spielstand. Beim Importieren wird dein aktueller Fortschritt ersetzt."
+
+#: scripts/components/popups/AboutPopup.tsx
+msgid "Export Save"
+msgstr "Spielstand exportieren"
+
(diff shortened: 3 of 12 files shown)


experintellia:
## Code review

Reviewed against the codebase (verified `applyDelta`'s guard chain, `_applyPeerDelta`, and `_mkDelta`). CI is fully green (typecheck, Biome, vitest, Playwright, build).

### Strengths
- **Security model is correct and verified.** The `importSave` reducer only runs for own-addr deltas because `applyDelta`'s Guard 3 (`scripts/state.ts:977`) returns before dispatch on foreign deltas. The reducer additionally re-overrides `addr`/`peers` from live state, so even a hand-crafted save embedding those fields can't hijack identity or wipe the leaderboard. Defense in depth, with a test covering the foreign-delta case.
- **Export field selection is compiler-enforced, not just convention.** `SaveState = Omit<LocalState, 'addr' | 'peers'>` forces `buildSaveState` to enumerate every other field — adding a future `LocalState` field is a *typecheck failure*, not a silent leak. Stronger than the "allow-list" framing in the description.
- **Monotonic clock is respected.** The reducer's `last_seen_ts: Math.max(...)` composes correctly with Guard 4's pre-advance; a stale imported timestamp can't rewind. Tested.
- **Durability-before-reload is handled.** Awaiting the `sendUpdate` thenable before reporting `{ok:true}` (so the popup's reload doesn't race the persistence write) is a subtle, real bug that was correctly anticipated — and the handler test asserts it.
- **Thorough tests** (round-trip, every `parseSaveFile` rejection branch, cold-start replay, peer-row refresh, partial/old saves).

### Issues worth considering

1. **Full snapshot is broadcast + persisted to every peer, forever (design concern).** The `importSave` delta carries the entire `state` snapshot (nodes, queues, etc.), and webxdc updates are append-only and replayed on every peer's device. Only `game_values` is consumed by `_applyPeerDelta` (`scripts/state.ts:911`); the rest is dead weight that permanently grows every participant's update log on each import. Consider trimming the broadcast payload to what the peer aggregator reads (`game_values` + the `info` message) and keeping the full snapshot local, if the replay/import path allows it. At minimum, worth flagging as a known tradeoff.

2. **Export button shows but silently no-ops where `sendToChat` is unavailable.** Visibility is gated only on `remoteApi.exportSave` existing (always true), while the real `webxdc.sendToChat` feature-detection lives inside `exportSave`, returning `{ok:false}` with no UI feedback (the result is fire-and-forget `void`). Import handles the parallel case correctly via the `'unavailable'` error. Either surface an error on export failure, or gate button visibility on the real capability.

3. **Importing player's own leaderboard `display_name` won't refresh.** `_applyPeerDelta` only updates `peer.display_name` on the `setDisplayName` op (`scripts/state.ts:920`). After importing a save with a different name, local `state.display_name` changes but the peer row's name doesn't. Probably acceptable, but worth a conscious decision.

### Nits
- Broadcasting the player's full state as plaintext JSON into a chat on export is intended (user-initiated backup), but worth a one-line note in the UI prose that it's plaintext/unencrypted.
- The `as unknown as void | Promise<void>` cast around `sendUpdate` is well-commented; no objection.

### Verdict
**Approve with minor follow-ups.** The correctness- and security-critical paths are sound and well-tested. Item #1 (full-snapshot gossip) is the only thing I'd want an explicit decision on before merge; #2 and #3 are polish that could land later.

---

Issue #275, opened 2026-05-12
Missing item descriptions: 134 game items use editorial placeholder text


experintellia:
In `data/ruleset_3.en.json`, **134 game items** carry the editorial-placeholder string "This text part of this beautiful game item is currently being reviewed by our amazing editorial team. It will become available here shortly. Thanks for your understanding." instead of real copy.

Totals:
- **152** placeholder strings across **134** items
- **98** missing `requirements_text` fields
- **54** missing `description` fields
- Affected: 11 perps (clients/database), 123 tokens

The German ruleset (`ruleset_3.de.json`) does not contain the placeholder — these gaps are English-only.

### Perps (11)

| ID | Title | Missing fields |
|---|---|---|
| `client002` | Health Insurance | requirements_text |
| `client005` | City Council | requirements_text |
| `client006` | Star Mart | requirements_text |
| `client007` | Car company | requirements_text |
| `client008` | Edu Inc | requirements_text |
| `client010` | Toys ´R´ Nice | requirements_text |
| `client011` | Major Label | requirements_text |
| `client014` | Bank | requirements_text |
| `client016` | Company.Com | description, requirements_text |
| `client019` | National Rental Housing | requirements_text |
| `database001` | Database | description |

### Tokens (123)

| ID | Title | Missing fields |
|---|---|---|
| `supertoken002` | Health Prognosis | requirements_text |
| `supertoken014` | Buying Habits | requirements_text |
| `token001` | First name | requirements_text |
| `token002` | Last name | requirements_text |
| `token003` | Address | requirements_text |
| `token004` | Zip Code | requirements_text |
| `token005` | Phone number | requirements_text |
| `token006` | Date of birth | requirements_text |
| `token007` | Sex | requirements_text |
| `token008` | Email | requirements_text |
| `token009` | Personnel Records | requirements_text |
| `token011` | Education | requirements_text |
| `token012` | Income | requirements_text |
| `token013` | Debts | requirements_text |
| `token014` | Weight | requirements_text |
| `token015` | Height | requirements_text |
| `token017` | Medical records | requirements_text |
| `token018` | Chronic illnesses | requirements_text |
| `token019` | Tanning booth visits | requirements_text |
| `token020` | Age at first sexual encounter | requirements_text |
| `token021` | Sexual orientation | requirements_text |
| `token022` | Relationship status | requirements_text |
| `token023` | Party photos | requirements_text |
| `token024` | Trustworthy or unreliable? | requirements_text |
[...]

PR #346, opened 2026-06-15
Simplify ruleset language handling — extract translatable strings into per-locale maps

experintellia:
## Summary

First step toward separating translations from game logic in the ruleset JSON files (related to #275).

This PR adds a one-time verification test that proves `ruleset_3.en.json` and `ruleset_3.de.json` are structurally identical — same keys, same nesting, same types, same array lengths — with only string leaf values allowed to differ (actual translations).

## Changes

- **tests/build/ruleset-structural-parity.test.js** — recursive walk comparing both files in parallel. Asserts:
  - Same top-level keys
  - Same nested keys at every level
  - Same types for all values (string vs number vs object vs array vs null)
  - Same array lengths
  - Non-string leaves (numbers, booleans, null) are identical
  - At least one known translation field actually differs (not a copy)

Once this is merged, Phases 1-4 will extract the translatable strings into a flat key→string map per locale and merge at runtime to eliminate duplication in the .xdc archive.

Co-authored-by: opencode big-pickle


experintellia:
## Phase 1+2 now included

**Phase 1** — Extraction script + generated files:
- `tools/extract-ruleset-i18n.mjs` walks both files in parallel, builds `ruleset_base.json` with `$t:path` markers, and per-locale flat string maps
- `data/ruleset_base.json` — 4302 extracted strings, all non-string data inline
- `i18n/ruleset.en.json` / `i18n/ruleset.de.json` — flat `{ "path": "text" }` maps

**Phase 2** — Runtime merge + verification:
- `scripts/inject-translations.ts` — pure function that walks base, resolves `$t:` markers against locale map
- `LocalEngine.ts` — `_getRuleset()` now merges base+strings at runtime instead of loading a pre-built full copy
- `tests/build/ruleset-i18n-merge.test.js` — lossless roundtrip proved: merged output deep-equals originals

### Build size impact (next step)
Currently both old files (`ruleset_3.en.json` / `ruleset_3.de.json`) still ship in the zip alongside the new base + locale files. Phase 3 will delete the old duplicates, which is where the real zip size savings come from — one base file + two small locale maps instead of two full 500KB+ copies.

experintellia:
## Phase 3 — Cleanup complete

Deleted old vendored files (`ruleset_3.en.json` / `ruleset_3.de.json`) and the one-time proof tests. Updated `type-settings-validation.test.js` to load base+locale maps and merge instead of reading old files directly. Build tests now verify new filenames in `dist/`.

The runtime merge in `_getRuleset()` has been proven lossless (Phase 2), and the old files were removed. The zip now stores one base file + two small locale maps instead of two ~600KB full copies.

### File count
- **Added**: 7 files (base, locale maps, inject-translations.ts, extraction script, test)
- **Deleted**: 6 files (old rulesets + 2 one-time proof tests)
- **Modified**: 5 files (LocalEngine.ts, UPSTREAM.txt, LICENSE-ASSETS.txt, build test, type-settings-validation test)

